PR #2, opened 2021-06-01
fix: fixes governance voting from front-end and adds add-contract method to DAO to introduce new contracts

philipdesmedt:


clarity/tests/arkadiko-dao_test.ts
```diff
@@ -0,0 +1,41 @@
...
+        types.principal("STSTW15D618BSZQB85R058DS46THH86YQQY6XCB7"),
+        types.principal("STSTW15D618BSZQB85R058DS46THH86YQQY6XCB7.arkadiko-mock-stacker-v1-1")
+      ], deployer.address)
+    ]);
+
+    block.receipts[0].result.expectOk().expectBool(true);
+  }
+});
+
+Clarinet.test({
```
nieldeckx, line 26: Q: I wrongly assumed the current set-contract-address could overwrite contracts. If we deploy an update of the staking contracts, the old ones need to be removed from the DAO. Otherwise both old and new contracts will be able to mint tokens. We could now test this with the mock-stacker.

clarity/contracts/arkadiko-dao.clar
```diff
@@ -112,6 +112,22 @@
   )
 )
 
+;; adds a new contract, only new ones allowed
+(define-public (add-contract-address (name (string-ascii 256)) (address principal) (qualified-name principal))
+  (begin
+    (asserts! (is-eq tx-sender (var-get dao-owner)) (err ERR-NOT-AUTHORIZED))
```
nieldeckx, line 118: REQ: By putting this method in the DAO, the protocol code become centralised. We could undo governance changes and deploy malicious contracts at any time. The DAO can not change and we need to be the 'dao-owner' to perform external tasks, so platform can not become fully decentralised?

My suggestion would be to put this feature in the governance contract. The 'dao-owner' can call the method on governance, and governance can call the existing 'set-contract-address' on the DAO.

clarity/tests/arkadiko-dao_test.ts
philipdesmedt: @nieldeckx The current `set-contract-address` can overwrite map values (not specific smart contracts?). Look at the difference of `map-set` vs `map-insert`. Or am I missing something?

clarity/contracts/arkadiko-dao.clar
philipdesmedt: @nieldeckx yes, it's true that this is a bit more centralised and the people who own the `dao-owner` address will be able to add new contracts to the protocol.

However, they cannot just overwrite other contract keys & values in the map. Look at the difference between `map-set`, which allows overwriting keys, vs `map-insert`, which returns false and does not set the new key/value pair when it already exists. So this method is addition only, no?

clarity/tests/arkadiko-dao_test.ts
nieldeckx: Ah yes, you are right.

clarity/contracts/arkadiko-dao.clar
nieldeckx: Ok, it's true the method can only be used to add new contracts but still provides a superpower which can not be removed later.

philipdesmedt: @nieldeckx ya - agree. 

I can move it to the governance contract, and keep the functionality the same, with the purpose that the governance contract can be replaced (and thus the functionality removed)?

philipdesmedt: @nieldeckx updated with your feedback! what do you think?

nieldeckx: @philipdesmedt perfect!

PR #6, opened 2021-06-03
feat: basic swap functionality

philipdesmedt:


nieldeckx:
@philipdesmedt changes made yesterday and today:
- fixed issue when removing liquidity
- fixed LP fee issue (swap function amounts now have 6 digits instead of 2)
- default fee-to-address is now the DAO payout address
- added tests and test models

clarity/contracts/arkadiko-stake-registry-v1-1.clar
```diff
@@ -117,15 +117,37 @@
 
 ;; Initialize the contract
 (begin
-  ;; Add initial contracts
+  ;; DIKO pool
   (map-set pools-data-map
     { pool: 'STSTW15D618BSZQB85R058DS46THH86YQQY6XCB7.arkadiko-stake-pool-diko-v1-1 }
     {
       name: "Diko",
```
philipdesmedt, line 124: Old code but can we name this "DIKO" - all caps

clarity/contracts/arkadiko-stake-registry-v1-1.clar
```diff
@@ -117,15 +117,37 @@
...
+  )
+  ;; DIKO-xUSD LP
+  (map-set pools-data-map
+    { pool: 'STSTW15D618BSZQB85R058DS46THH86YQQY6XCB7.arkadiko-stake-pool-diko-xusd-v1-1 }
+    {
+      name: "DIKO-xUSD LP",
+      active: true,
+      activated-block: block-height,
+      deactivated-block: u0,
+      rewards-percentage: u400000 ;; 40%
```
philipdesmedt, line 139: Q: How did you decide 40% for each?

Are we more interested in liquidity on DIKO/xUSD rather than wSTX/xUSD? Maybe put the needle slightly to DIKO with a 20 / 50 / 30? or 20 / 45 / 35?

clarity/contracts/arkadiko-swap-v1-1.clar
```diff
@@ -0,0 +1,435 @@
...
+  (let (
+    (token-x (contract-of token-x-trait))
+    (token-y (contract-of token-y-trait))
+    (pair (unwrap-panic (map-get? pairs-data-map { token-x: token-x, token-y: token-y })))
+    (balance-x (get balance-x pair))
+    (balance-y (get balance-y pair))
+    (sender tx-sender)
+    (dx-with-fees (/ (* u997 dx) u1000)) ;; 0.3% fee for LPs
+    (dy (/ (* balance-y dx-with-fees) (+ balance-x dx-with-fees)))
+    (fee (/ (* u5 dx) u10000)) ;; 0.05% fee for protocol
```
philipdesmedt, line 297: @nieldeckx Q: are you taking 0.35% now? What uniswap does is 0.25 + 0.05, I think that makes more sense?

nieldeckx: No, the user pays 0.3% in fees out of which the protocol can take 0.05%.

philipdesmedt: Ok 👍

clarity/contracts/arkadiko-stake-registry-v1-1.clar
nieldeckx: I could not decide ;) 

I'm a bit worried that the rewards to provide wSTX/xUSD are not high enough. Seems much better to just put STX in a vault. You get pox rewards, DIKO rewards and can mint xUSD. With the DIKO rewards and minted xUSD you can start providing liquidity. You again earn DIKO which can be used to add to your LP position. In this scenario you keep full STX exposure while earning pox rewards, so the end result for an STX holder is always positive. 

But if you want to provide wSTX/xUSD liquidity you need to put some STX in the pool directly. You will suffer impermanent loss and won't be able to stack. In this scenario an STX holder could end up negative compared to just stacking STX, so this is a higher risk. This pool might be important for xUSD as it allows arbitrage so helps stabilising xUSD value?

philipdesmedt: @nieldeckx good points - didn't think of that. Maybe we should do it the other way around then even, more rewards for wSTX/xUSD :)

20 / 30 / 50 or 20 / 35 / 45. I'll CC in ozzy through chat

Macxim:
I added some UI/UX improvements. It's just a start, I'm sure it can still be improved though.

#### Swap tokens

|Before|After|
|---|---|
|b|a|

#### Transactions settings

Screen Shot 2021-06-07 at 11 11 24 AM


#### Token list

Tokens' logo are temporary, nothing is set is stone, let's iterate on this:
Screen Shot 2021-06-07 at 11 11 38 AM

philipdesmedt:
👏

Macxim:
@philipdesmedt We need to prevent unwanted characters as well on both token inputs and slippage tolerance, only to allow `0-9` and `.`.

PR #8, opened 2021-06-07
tweak: add several swap improvements

philipdesmedt:


web/components/swap.tsx
```diff
@@ -85,6 +85,7 @@ export const Swap: React.FC = () => {
           const balanceY = json4['value']['value']['value']['balance-y'].value;
           const basePrice = (balanceX / balanceY).toFixed(2);
```
Macxim, line 86: `toFixed` retuns a string.

```suggestion
          const basePrice = Number((balanceX / balanceY).toFixed(2));
```

Same on L86

PR #15, opened 2021-06-10
Clean up testnet modal UI

Macxim:
|Before|After|
|---|---|
|Screen Shot 2021-06-10 at 12 12 41 PM|Screen Shot 2021-06-10 at 12 12 24 PM|


philipdesmedt:
🙇 @Macxim

PR #20, opened 2021-06-13
Mint: UI improvements

Macxim:
|Before|After|
|---|---|
|bef|aft|


philipdesmedt:
🙇‍♂️

Macxim:
What should we do about xUSD last price?

philipdesmedt:
@Macxim should be fixed at $1, for now at least. Later when we go to mainnet, we will probably get a real-time price from the swap

philipdesmedt (approved):
Looks great man! Thanks for fixing my shitty code :)

PR #24, opened 2021-06-16
feat: add link to use max balance/mint

hstove:
I wanted there to be a "max" button, so I used this as an excuse to get the project setup and contribute 🙂

image

I totally understand if you don't want to merge this, please feel free to just close it if you'd rather take a different approach.

Some of the changes are just from VSCode running ESLint. I had to make some changes to get ESLint to work in my workspace, and I can push those if you're interested.

### DIFF
--- a/web/common/context.ts
+++ b/web/common/context.ts
@@ -10,6 +10,8 @@ interface UserBalance {
   stdiko: number;
 }
 
+export type UserBalanceKeys = keyof UserBalance;
+
 export interface CollateralTypeProps {
   name: string;
   token: string;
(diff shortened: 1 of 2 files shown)


philipdesmedt:
@hstove awesome, thanks for the contribution! The front-end code can definitely use some love (both cleaning up & improving UX haha).

One little issue with this: you collateralise STX, but if you collateralise your "max balance", you don't have STX anymore to pay for the tx fees. Any idea how to fix this? A straightforward solution would be to check if the collateral is the STX token, and just subtract X (enough to pay for fees)?

hstove:
Good point! You can definitely do the thing you mentioned. Right now since fees are fixed, this should be pretty straightforward

hstove:
I've added a feature to withdraw `0.0004` STX from the balance when using STX, which is the current fee when opening a STX vault

philipdesmedt:
@hstove thanks! will review & merge tomorrow :)

philipdesmedt (approved):
Works! 

Thanks for this. Didn't know about `useCallback` :)

PR #71, opened 2021-06-21
feat: add more dynamic traits

philipdesmedt:


nieldeckx (changes requested):
The 'mint' function in freddie does not check if the given oracle is active in the DAO. Should it?

philipdesmedt:
@nieldeckx added, thx!

nieldeckx:
Cool! LGTM

PR #83, opened 2021-06-23
fix: respect slippage for minimum received only, not price

philipdesmedt:


Macxim (commented):
Works fine 👍 well done.

PR #85, opened 2021-06-23
Chore: update clarinet 0.13.0

lgalabru:
There was an issue in the way the `expectUint` was written, that was fixed with clarinet v0.13.0.
The following behavior would have been observed in prior versions:

```
"u100".expectUint(1); // pass
"u100".expectUint(10); // pass
"u100".expectUint(100); // pass
"u100".expectUint(1000); // fail
```

This PR is updating the version being imported, and fixing the broken expectations (which looks like broken tests to me, vs broken contracts).

### DIFF
--- a/clarity/tests/arkadiko-auction-engine-v1-1_test.ts
+++ b/clarity/tests/arkadiko-auction-engine-v1-1_test.ts
@@ -4,7 +4,7 @@ import {
   Clarinet,
   Tx,
   types,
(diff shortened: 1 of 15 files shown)


philipdesmedt:
@lgalabru thanks for this! sneaky bug

@nieldeckx seems important to update your local clarinet dist. You need to install this manually for now like this

```
git clone https://github.com/hirosystems/clarinet.git
cargo install --path .
```

PR #86, opened 2021-06-24
fix: refactor websocket listener

philipdesmedt:


philipdesmedt:
@nieldeckx @Macxim this should be ready for review. I rewrote the websocket connection cause it was a bad implementation previously.

Previously I would try to create a new WS connection on each page where a TX happens. Now 1 stream is opened on login/initial page load and it listens to all transactions.

philipdesmedt:
@Macxim could you fix the close icon in the popup?
image

Macxim:
Will do!

Issue #60, opened 2021-06-20
Auction won shows 1000 xUSD bid instead of actual 2 xUSD


314159265359879:
https://explorer.stacks.co/txid/0x47eefc70e6a6b3802d4effc3c2fc51da1d942dbc4cf7724b558ef1aced329afb?chain=testnet
Transaction for lot 21 shows a 1000 xUSD whereas I made 2 xUSD bid.

![image](https://user-images.githubusercontent.com/33360391/122687666-e734a580-d217-11eb-8e58-7dbd62633ff4.png)

I checked if I accidentally did another bid with 1000 xUSD, but I do not see it on my transactions list. I do see 2 xUSD bids before and after for lot [19 (nonce 26)](https://explorer.stacks.co/txid/0x684c12046cdca1746c124369f138d3126b953aff477f8ee38c7483bbd44c46d7?chain=testnet) and [23 (nonce 28)](https://explorer.stacks.co/txid/0xd7753126cb7285fcc63b37492d84473b485299ebf2612d64ae49be42c0aed5ef?chain=testnet), I forgot 22 (nonce 43) and did that later and that is why it is skipped in this [list too](https://user-images.githubusercontent.com/33360391/122687924-5f4f9b00-d219-11eb-93c8-23b71ce4b04c.png).

Issue #58, opened 2021-06-20
Same lot shows twice


nieldeckx:
<img width="1141" alt="Screenshot 2021-06-20 at 17 48 07" src="https://user-images.githubusercontent.com/5592676/122680489-b3458a00-d1ef-11eb-9125-7030a1e26810.png">

314159265359879:
![image](https://user-images.githubusercontent.com/33360391/122687018-82c41700-d214-11eb-9791-3e1f0d4db228.png)
Did some more research on this. 
When I pay the full price twice, one transaction simply fails ([1](https://explorer.stacks.co/txid/0xf19152482e19654a202d99aa1df2d1cdd8bcae87ffde40b7df6d868952bdc9aa?chain=testnet), [2](https://explorer.stacks.co/txid/0xc20137b93fa16110ec42d069208e1089cb475c8e0431c5f9e2b985112ad0bd66?chain=testnet)). That is what I would expect. 
When I bid 2 xUSD first and 117 xUSD (full price) second... that is when this happens. Showing same lot twice.
[Bid 1 2 xUSD](https://explorer.stacks.co/txid/0x68f1d7fd08cbcc3c5fd09138b94095f9ae63b2ccd84d57dcfd1c3e4e1e2fac9a?chain=testnet) 
[Bid 2 117 xUSD](https://explorer.stacks.co/txid/0x7eabf2e5e2ec72f1a240e49709a09e9fd60e6abbb2d3e452519304afae13db2e?chain=testnet) 

Since in this case I am bidding 119 xUSD on the lot it would make more sense to add those numbers and have one instance with 119. First time this happened to me I just made the mistake of bidding twice on the same lot, that is less likely to happen on mainnet.
I can see people making a mistake and then outbidding themselves by bidding again. It would make sense if that is "added" to the bid for the lot, right?

PR #88, opened 2021-06-24
fix: fix several auction issues

philipdesmedt:


philipdesmedt:
@nieldeckx this cleans up the auction ID list and should fix most auction issues afaik (maybe #58 and #60).. not sure cause it's hard to replicate

philipdesmedt:
Also implemented #54

clarity/contracts/arkadiko-auction-engine-v1-1.clar
```diff
@@ -554,6 +578,7 @@
 
 (define-private (extend-auction (auction-id uint))
   (let ((auction (get-auction-by-id auction-id)))
+    (print "EXTENDING AUCTION!!!")
```
nieldeckx, line 581: PP: DO NOT USE CAPITALS!!! ;)

clarity/contracts/arkadiko-auction-engine-v1-1.clar
```diff
@@ -439,9 +447,21 @@
       )
       (try! (contract-call? vault-manager redeem-auction-collateral ft token-string reserve (get collateral-amount last-bid) tx-sender))
     )
+    (if (< (get xusd last-bid) (var-get lot-size))
+      (begin
+        (map-set auctions
+          { id: auction-id }
+          (merge auction {
+            lots-sold: (+ u1 (get lots-sold auction)),
```
nieldeckx, line 455: Q: lots-sold is already updated on 'register-bid', why again when trying to redeem?

clarity/contracts/arkadiko-auction-engine-v1-1.clar
```diff
@@ -63,6 +63,7 @@
 (define-data-var auction-ids (list 1500 uint) (list u0))
 (define-data-var lot-size uint u1000000000) ;; 1000 xUSD
 (define-data-var auction-engine-shutdown-activated bool false)
+(define-data-var removing-auction-id uint u0)
```
nieldeckx, line 66: Q: why do we need to store this as a var on the SC?

clarity/contracts/arkadiko-auction-engine-v1-1.clar
philipdesmedt: Leftover from debugging, thx :D

clarity/contracts/arkadiko-auction-engine-v1-1.clar
philipdesmedt: In `register-bid` it is only updated if you bid over lot size (1000 USDA).. in here it is updated when you redeem (unless it was already set as sold since we check that bid < 1000 USDA)

clarity/contracts/arkadiko-auction-engine-v1-1.clar
philipdesmedt: else we don't know which auction we're removing in the `filter` method on line 604

nieldeckx:
Alright cool, thx for explaining.

PR #89, opened 2021-06-24
tweak: tighten collateralization on collateral types

philipdesmedt:


clarity/contracts/arkadiko-freddie-v1-1.clar
```diff
@@ -295,7 +295,7 @@
     )
     (asserts! (is-eq (contract-of coll-type) (unwrap-panic (contract-call? .arkadiko-dao get-qualified-name-by-name "collateral-types"))) (err ERR-NOT-AUTHORIZED))
     (asserts! (is-eq (contract-of oracle) (unwrap-panic (contract-call? .arkadiko-dao get-qualified-name-by-name "oracle"))) (err ERR-NOT-AUTHORIZED))
-    (asserts! (>= ratio (get liquidation-ratio collateral-type-object)) (err ERR-INSUFFICIENT-COLLATERAL))
+    (asserts! (>= ratio (get collateral-to-debt-ratio collateral-type-object)) (err ERR-INSUFFICIENT-COLLATERAL))
```
nieldeckx, line 298: Safer indeed :) 
Wondering if we should have a similar check in the 'mint' function on Freddie?

nieldeckx (commented):
This is better indeed. Just one small question.

clarity/contracts/arkadiko-freddie-v1-1.clar
philipdesmedt: Yeah - good point. I will add this.

PR #101, opened 2021-06-28
feat: add liquidity through the UI

Macxim:
Mostly static at the moment but here it is:

As usual, nothing is set in stone, feedback welcome!


philipdesmedt:
Don't have much to comment on this, maybe a few small things:

- At the bottom, you can see "Your pool tokens" & share.. would these be the current ones that you already have added (before potentially adding more) or based on the ones you're selecting?
- Pooled xUSD & Pooled DIKO is the global current pool I guess?

Macxim:
Ah yeah, sorry it's not clear in the screenshot.
- The block with `Your pool tokens & share` would be filled in as the user fills the tokens input
- This displays the pair that is currently selected

nieldeckx:
Looks good!

nieldeckx (approved):

Macxim:
Do we also need a link at the top leading to the current open positions?

Macxim:
We need to change the copy for the tooltips in `Pool tokens` and `Pool share`.

philipdesmedt:
@Macxim we can do the liquidity positions in another PR, but seems low priority right now.

PR #106, opened 2021-06-29
Rename xUSD to USDA

nieldeckx:
Changed file names as well. Tests and website still working.


philipdesmedt (approved):
Great work! Gonna merge this.

We need to adjust documentation on gitbook as well

PR #108, opened 2021-06-30
Fix link in landing

victorcoder:
The link was pointing the old repo


philipdesmedt:
thx @victorcoder! good find

PR #109, opened 2021-07-01
fix: calculate get-total-staked as total amount staked, not just for user

philipdesmedt:
@nieldeckx I think you implemented this for the user, but that is not in line with how the other pool contracts calculate the `get-total-staked` amount? Also it crashes in the front-end now..

Can you elaborate if this is correct or not?

### DIFF
--- a/clarity/contracts/arkadiko-stake-pool-diko-v1-1.clar
+++ b/clarity/contracts/arkadiko-stake-pool-diko-v1-1.clar
@@ -60,16 +60,8 @@
     ;; Total DIKO
     (total-diko-supply (+ diko-supply rewards-to-add))
 
-    ;; Total stDIKO supply
-    (calculated-stdiko-supply
-      (if (is-eq u0 stdiko-supply)
-        (unwrap-panic (contract-call? .stdiko-token get-total-supply))
-        stdiko-supply
-      )
-    )
-
     ;; User stDIKO percentage
-    (stdiko-percentage (/ (* amount u1000000) calculated-stdiko-supply))
+    (stdiko-percentage (/ (* amount u1000000) stdiko-supply))
 
     ;; Amount of DIKO the user will receive
     (diko-to-receive (/ (* stdiko-percentage total-diko-supply) u1000000))
@@ -135,7 +127,7 @@
 
     (let (
       ;; Amount of DIKO the user will receive
-      (diko-to-receive (unwrap-panic (diko-for-stdiko registry-trait amount u0)))
+      (diko-to-receive (unwrap-panic (diko-for-stdiko registry-trait amount (unwrap-panic (contract-call? .stdiko-token get-total-supply)))))
     )
       ;; Burn stDIKO 
       (try! (contract-call? .arkadiko-dao burn-token .stdiko-token amount staker))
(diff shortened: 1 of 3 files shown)


nieldeckx:
@philipdesmedt Oops sorry. You are right, this method should just return the total DIKO in the pool contract.

I guess we also need a method `(define-read-only (get-stake-of (staker principal))` on this pool to show in the frontend? You can use the removed code for this?

philipdesmedt:
@nieldeckx OK! will adapt in this PR

philipdesmedt:
@nieldeckx can you look at the code? I am not convinced it's correct now that I'm testing it all on the front-end.. The staked amount always stays the same, while it should go up per block no?

It's been at 9K DIKO for an hour (And never changed since I staked)
image

nieldeckx:
@philipdesmedt That's probably because there was no other interaction with the pool? Auto-compounding happens automatically when a user stakes/unstakes or when calling "add-rewards-to-pool". Code looks good to me.

philipdesmedt:
@nieldeckx that'd be great yes! Important that it stays a read-only fn tho

nieldeckx:
@philipdesmedt Done.

philipdesmedt:
@nieldeckx awesome! will review, adapt the front-end if needed, and merge 👍

philipdesmedt:
@nieldeckx can you review this again? I had to refactor some small things in the smart contract as well, since the read count was over 30 (it was 32) meaning we were doing too many `contract-call` invocations etc. I couldn't really eliminate any so now we send the stdiko-supply from the front-end

clarity/contracts/arkadiko-stake-pool-diko-v1-1.clar
```diff
@@ -61,10 +61,15 @@
     (total-diko-supply (+ diko-supply rewards-to-add))
 
     ;; Total stDIKO supply
-    (stdiko-supply (unwrap-panic (contract-call? .stdiko-token get-total-supply)))
+    (calculated-stdiko-supply
```
nieldeckx, line 64: PP: Remove this logic where the parameter can be zero. On line 138, just pass the total supply instead of 0.

clarity/contracts/arkadiko-stake-pool-diko-v1-1.clar
```diff
@@ -169,14 +174,13 @@
...
-    (block-diff (- last-block-height (var-get last-reward-add-block)))
+    (last-block-info (get-last-block-height registry-trait))
+    (block-diff (- (get height last-block-info) (var-get last-reward-add-block)))
     (rewards-to-add (* rewards-per-block block-diff))
-    (deactivated-block (unwrap-panic (contract-call? registry-trait get-pool-deactivated-block .arkadiko-stake-pool-diko-v1-1)))
   )
     ;; Rewards to add can be 0 if called multiple times in same block
     ;; Do not mint if pool deactivated
-    (if (or (is-eq rewards-to-add u0) (not (is-eq deactivated-block u0)))
+    (if (or (is-eq rewards-to-add u0) (is-eq false (get pool-active last-block-info)))
```
nieldeckx, line 183: Better to be explicit indeed :)

nieldeckx (commented):
Great, just one minor comment

PR #111, opened 2021-07-01
Fix inputs fields where Max button would be pushed out of container

### LINKED ISSUE #110: Long tickers push Max button in inputs out of container
Screen Shot 2021-07-01 at 11 27 20 AM

Macxim:
Fix #110 

Now:

Screen Shot 2021-07-01 at 11 31 29 AM


philipdesmedt:
@Macxim nice!

PR #112, opened 2021-07-01
feat:add a single button to add/remove liquidity

Macxim:
<p align="center">
Screen Shot 2021-07-01 at 3 50 03 PM
</p>


nieldeckx:
I would not combine the "add/remove liquidity" button and the secondary information in the same light purple box. The information is linked to your current swap, while the new button is a different action. Something like this makes more sense to me:
Screenshot 2021-07-01 at 17 05 55

Macxim:
@nieldeckx 
Yes, you are absolutely right. 

We could also have a button group like this at the top: 

image

What do you think?

nieldeckx:
Yeah, that's probably better :)

Macxim:
Alright!

Macxim:
Since it's a link (that takes into account the current pair in the URL), it should look like so, and not like a button.
We can do the whole Pool overview and differentiation between Swap and Pool later.

Here it is, this should be enough for the short term and in the meantime we can focus on other things.

Screen Shot 2021-07-02 at 4 06 18 PM

PR #113, opened 2021-07-01
Upgrade deps and enable npm-only usage

victorcoder:
This PR is intended to make the web compatible with npm-only usage.

Nowadays npm is as full featured as yarn and most people tend to use only npm for development.

npm has a stronger dependency check, it was complaining of some unmet dependency requirements, so this PR also upgrades the deps and include the package-lock.json file for npm.


philipdesmedt:
@victorcoder very cool! 

can we still use `yarn` alongside `npm` or do you suggest we get rid of `yarn` altogether?

victorcoder:
I prefer to use npm, but could be used alongside, personally I will get rid of yarn.

philipdesmedt:
@victorcoder nice work! merging this

PR #116, opened 2021-07-04
line 211 and 212 were duplicates.

fiftyeightandeight:
Both check if pairs-data-map contains the pair.


philipdesmedt:
The current code is correct. We check if the pair (tokenX, tokenY) or (tokenY, tokenX) exists (note the X,Y vs Y,X). The pairs map takes an ID as key btw.. not a pair. 

Closing this :)

PR #122, opened 2021-07-08
feat: LP token staking

philipdesmedt:


philipdesmedt:
@Macxim any ideas on these stake/unstake/claim buttons?

image

nieldeckx:
Looks good, better to have APY calculations in FE. 

Proposal to visualise info and buttons:
Screenshot 2021-07-10 at 14 20 30

philipdesmedt:
@nieldeckx nice! 

@Macxim what do you think?

Macxim:
Mmmh, ideally actions should be grouped in the same place to avoid confusion, preferably on the far right for LTR locales. 

Basically it boils down to this:

- inline actions like we currently have (but 3 is already a lot IMO)
- hover actions: actions are hidden and only shown when hovering the row, reducing clutter
- menu actions: dedicated dropdown menu for each row where you put all the available actions

You can alternatively have a combination 1 and 3, that is keeping most important action inline and having secondary actions hidden behind a dropdown menu. 

More info here:

- https://ux.stackexchange.com/questions/26366/best-way-to-show-many-user-actions-for-table-rows
- http://www.jankoatwarpspeed.com/ultimate-guide-to-table-ui-patterns/

philipdesmedt:
@Macxim should I merge and you improve it later or you wanna fix some small things on this branch?

Macxim:
I guess you can merge. I'll have a look tomorrow or later this week. Good job! 💪

PR #125, opened 2021-07-10
fix: allow vault owner to unlock vault after PoX cycle as well

philipdesmedt:


nieldeckx (commented):
Q: Can the same be done for the method "release-stacked-stx"? If the necessary check are in place, there is no reason to only let the dao-owner call these methods?

philipdesmedt:
@nieldeckx I'll have to check this, not sure

nieldeckx:
> @nieldeckx I'll have to check this, not sure

I think we can allow users to call this method because the necessary checks are in place and it can only be called once per vault.

philipdesmedt:
@nieldeckx yep - true. I will probably close this PR in favor of the other one (auto payoff) since have a ton of refactorings going on there.

PR #126, opened 2021-07-10
fix: mint new DIKO when bad debt needs to be covered by a debt auction

philipdesmedt:
When bad debt had to be covered and STX collateral was not sufficient, we would previously take this from the foundation reserve.

That resulted in a systemic risk where tokens from the foundation could be drained. This is now changed and we mint new DIKO tokens.

### DIFF
--- a/clarity/contracts/arkadiko-dao.clar
+++ b/clarity/contracts/arkadiko-dao.clar
@@ -148,9 +148,13 @@
 
 ;; This method is called by the auction engine when more bad debt needs to be burned
 ;; but the vault collateral is not sufficient
-;; As a result, this method requests DIKO from the DAO ("foundation reserves")
-(define-public (request-diko-tokens (ft <ft-trait>) (collateral-amount uint))
-  (contract-call? ft transfer collateral-amount (var-get dao-owner) (as-contract (unwrap-panic (get-qualified-name-by-name "sip10-reserve"))) none)
+;; As a result, additional DIKO will be minted to cover bad debt
+(define-public (request-diko-tokens (collateral-amount uint))
+  (begin
+    (asserts! (is-eq (unwrap-panic (get-qualified-name-by-name "auction-engine")) contract-caller) ERR-NOT-AUTHORIZED)
+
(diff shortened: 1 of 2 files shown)


clarity/contracts/arkadiko-dao.clar
```diff
@@ -148,9 +148,13 @@
 
 ;; This method is called by the auction engine when more bad debt needs to be burned
 ;; but the vault collateral is not sufficient
-;; As a result, this method requests DIKO from the DAO ("foundation reserves")
-(define-public (request-diko-tokens (ft <ft-trait>) (collateral-amount uint))
-  (contract-call? ft transfer collateral-amount (var-get dao-owner) (as-contract (unwrap-panic (get-qualified-name-by-name "sip10-reserve"))) none)
+;; As a result, additional DIKO will be minted to cover bad debt
+(define-public (request-diko-tokens (collateral-amount uint))
```
nieldeckx, line 152: REQ: Is it possible to use the existing "mint-token" method on the DAO? We should avoid specific methods on the DAO as this contract can not be changed.

nieldeckx (changes requested):

clarity/contracts/arkadiko-dao.clar
philipdesmedt: @nieldeckx this is hard and requires a lot more changes since I would have to pass in a `dao-token` trait instance, which requires us to refactor the auction engine.

I don't see a problem here since neither the `arkadiko-token` contract nor this one (dao) can be changed... or am I missing something?

nieldeckx: @philipdesmedt Agree, not a real issue :) As the DAO can not change, I wanted to keep it as simple as possible but it's not worth the refactor to the auction engine.

philipdesmedt: @nieldeckx I guess this one method is fine since it is a protocol method that closely interacts with the mint function to create a liquidation mechanism. Likely this will never have to change

PR #127, opened 2021-07-10
feat: auto payoff a vault's loan through the AMM

philipdesmedt:


philipdesmedt:
@nieldeckx this is ready for a first review but requires the nightly build of `clarinet`. This is also why the tests on CI are currently failing, it needs a Clarinet upgrade.

nieldeckx (changes requested):
Couple of things before I continue testing:
- usda-token_tests is failing on L36 because wallet_3 has not USDA
- remove test statement on L56 in stacker (to fix tests)
- stacker test "payout two vaults" failing, I think because in the stacker contract on L240 there is not enough STX in the reserve?
- when stacking minimum is not reached, error code is now 18 instead of 191. Not sure why?
- in stacker test "payout two vaults" vault 1 has auto-payoff enabled, but at the end of the test it checks that the collateral increased. I thought with auto-payoff, the collateral would remain the same but the debt would decrease?

philipdesmedt:
@nieldeckx fixed the stacker test "payout two vaults"

The `u18` thrown is coming from the actual PoX contract and is the following: `(define-constant ERR_STACKING_INVALID_AMOUNT 18)`, see https://explorer.stacks.co/txid/0x41356e380d164c5233dd9388799a5508aae929ee1a7e6ea0c18f5359ce7b8c33?chain=mainnet

I thought it made more sense to forward all the errors from the actual PoX contract instead of abstracting away the errors, will fix the test

philipdesmedt:
@nieldeckx one thing to consider in the tests is what happens:

- slippage
- transfer failed
- excess USDA

Haven't really thought of all the edge cases but would be great if you could (dis)cover these

nieldeckx:
@philipdesmedt Cool!

Before going into the edge cases, I added tests which cover basic funtionality but are failing.

1) Not able to pay out vault which has auto-payoff on. STX reserve has no STX to swap to USDA. Do you know how we can transfer STX from wallet to reserve in unit tests?

2) When creating a vault with stack-pox off, it will still count towards the tokens-to-stack on the STX reserve. Because the "collateralize-and-mint" function on the reserve does not take into account the auto-payoff setting. I guess we can just add a parameter on this function?

philipdesmedt:
@nieldeckx 

1. Does the STX-reserve contract need the STX? I think the swap contract needs to have liquidity, no? You can check the swap tests to add liquidity? Or what am I missing?
2. I see - we could add an additional parameter, but the `auto-payoff` parameter already signals this no? Just needs to be taken into account in the relevant method?

nieldeckx:
@philipdesmedt Pushed 2 fixes we discussed. 

Additionally:
- As we keep stacking even when all debt is paid, we should not call "enable-vault-withdrawals". This failed and prevented us from paying back all debt for a vault.
- Excess USDA is send to the vault owner's wallet

philipdesmedt:
@nieldeckx right, good point on the first one.

philipdesmedt:
@nieldeckx let me know when you want me to review btw.. can write some additional tests as well

nieldeckx:
Fixed issues when stacking is revoked for vault on auto-payoff.
@philipdesmedt please review. Other scenario's you can think of that we should test?

philipdesmedt:
@nieldeckx cool - will have a look, test and review.. will comment here if there's edge cases and other things I think of

philipdesmedt:
@nieldeckx @Macxim @victorcoder this is good to merge for a v1 - will iterate another time in the next two weeks.

A few things have changed through this PR:

1. You have to use `clarinet integrate` instead of `./manage.sh mocknet up/down` to run a local environment. Clarinet will automatically run everything for you through Docker instances as well, such as a bitcoin blockchain and an explorer, which allows you to test PoX locally with "real" bitcoin. If you haven't installed clarinet yet, now is a good time :)
2. You will see the addresses changed. `ST1PQHQKV0RJXZFY1DGX8MNSNYVE3VGZJSRTPGZGM` is now the deployer. Make sure to update in your .env files in the `web` and `scripts` folder (also the private key, check `Devnet.toml`).
3. The port for the web app has changed to 9000 because clarinet uses 3000 to run the bitcoin explorer
4. The port for the Stacks API node is now 20080 instead of 3999

Text me on discord if you have any questions

PR #131, opened 2021-07-12
Add actions dropdown menu for Staking

Macxim:
https://user-images.githubusercontent.com/1909957/125280825-6c8a1200-e30d-11eb-9f46-b5eecd3491ea.mov


nieldeckx:
Nice. Is it possible to disable a button in the menu? For example:
- Disable unstake button if no funds staked
- Disable claim button if no rewards to claim

Just a question / nice to have. Not critical.

Macxim:
@nieldeckx 

I think we could, yes.

However, like for most disabled buttons, it could create confusion instead of bringing value:
- why is this button disabled?
- why can't I click on this?
- what am I missing? what do I have to do to enable this?

IMO, if no funds are staked or there are no rewards to claim, those buttons shouldn't be shown.

philipdesmedt:
@Macxim @nieldeckx I can hide the dropdown buttons if we want to do that.

Let me know. I think it's a good solution.

Macxim:
@philipdesmedt Yes, go ahead.

PR #137, opened 2021-07-14
Fix width of switcher so that tokens are all readable

Macxim:
|Before|After|
|---|---|
|Screen Shot 2021-07-14 at 10 43 07 AM|Screen Shot 2021-07-14 at 10 42 25 AM|


nieldeckx (approved):
Perfect :)

PR #144, opened 2021-07-18
User can now pay back all debt while stacking

nieldeckx:
- A user can pay back all debt when they want (vault not auto closed)
- A user can close a vault that has no debt left (and is not stacking STX)
- When closing a vault with debt, it will be paid back automatically


philipdesmedt (approved):
Good.

We now need to update the UX with your new design and make sure you can't close the vault unless you unstacked your STX tokens (and current PoX cycle has ended).

PR #149, opened 2021-07-23
fix: fix potential vulnerability in ft-transfer?

philipdesmedt:


nieldeckx (approved):
Can you give a bit more context? Just trying to understand.

https://docs.stacks.co/references/language-functions#as-contract
"The as-contract function switches the current context's tx-sender value to the contract's principal"

In the following change for example, I don't think it matters who is the tx-sender of the internal function call as it's never used in that function. It does matter for the parameter.
`(contract-call? .usda-token transfer usda-amount (as-contract tx-sender) (contract-call? .arkadiko-dao get-payout-address) none)`
`(as-contract (contract-call? .usda-token transfer usda-amount (as-contract tx-sender) (contract-call? .arkadiko-dao get-payout-address) none))`

Anyway, based on the docs and tests I do think this is fine :)

philipdesmedt:
@nieldeckx it does matter

The first sets the `sender` address to the contracts address, but does not execute the `transfer` function **as** the contract. The last part is crucial. The second statement switches the actual context as well (i.e. truly sending "as" the contract).. AFAIK

nieldeckx:
Strange that everything was working before this change :P

philipdesmedt:
@nieldeckx well it worked cause you could spend funds from any address without having to be in the context of that address... now not anymore

PR #153, opened 2021-07-26
tweak: pay stability fee on minting USDA if enough balance

### LINKED ISSUE #38: Minting extra xUSD also tries to pay stability fees in same TX which fails if xUSD balance is 0
image

philipdesmedt:
Fixes #38 

There's two options:

1. This approach
2. Remove the paying of stability fee in the mint method

### DIFF
--- a/clarity/contracts/arkadiko-freddie-v1-1.clar
+++ b/clarity/contracts/arkadiko-freddie-v1-1.clar
@@ -491,6 +491,7 @@
       (err ERR-MAXIMUM-DEBT-REACHED)
     )
 
+    (try! (accrue-stability-fee vault-id coll-type))
     (try! (contract-call? reserve mint
         (get collateral-token vault)
         (get owner vault)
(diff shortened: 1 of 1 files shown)


nieldeckx (commented):
I prefer option 2 because:
- This is more transparent to the end user (otherwise we need to explain this extra logic in the UI)
- Our marketing page mentions the fee can be paid back whenever they want
- I get we want to push users to pay the stability fee asap. I think we can do so via the UI. In this implementation the user might not have enough USDA, or might just create a new vault if he does not want to pay the stability fee yet.

philipdesmedt:
@nieldeckx okay, I'll just remove paying the stability fee!

philipdesmedt:
@nieldeckx done

philipdesmedt:
@nieldeckx this PR is not done yet, I need to accrue the stability fees, since the interest based on the minted amount has changed, will fix this

philipdesmedt:
@nieldeckx done

nieldeckx:
Alright, looks good.

PR #154, opened 2021-07-27
fix: calculate staking APYs based on each staking pool

nieldeckx:
APY is now calculated for each pool.

### DIFF
--- a/web/components/stake.tsx
+++ b/web/components/stake.tsx
@@ -206,22 +206,22 @@ export const Stake = () => {
       if (totalDikoStaked === 0) { totalDikoStaked = 10 };
       const dikoPoolRewards = totalStakingRewardsYear1 * 0.1;
       const dikoApr = dikoPoolRewards / totalDikoStaked;
-      setApy(Number((100 * dikoApr / 100000).toFixed(2)));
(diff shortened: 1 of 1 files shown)


philipdesmedt (approved):
Code looks good. Will test this on the front-end and merge.

We will need some loading states, since we are waiting for >5 requests now, but I'll handle that

philipdesmedt:
@nieldeckx I divided all APY's by 10^5 since I think that was needed?

Also added loading state

nieldeckx:
Love the loading state.

@philipdesmedt I don't think you needed to divide the APY's again? When fetching the total tokens staked we already divide by 1000000. 

Simple example for DIKO pool:
- Total staking reward for year 1 = 23500000 * 0.1 = 2350000 (see const totalStakingRewardsYear1)
- User stakes 2350000 DIKO (see totalDikoStaked where we already divide by 1000000)
- APY = 2350000/2350000 = 1 = 100% (user indeed doubles his amount over 1 year)

philipdesmedt:
@nieldeckx not sure... the APYs were like 41700000% :P seems really high?

will double check before merging

philipdesmedt:
@nieldeckx you were right, rolled back the divisions. Gonna merge

PR #155, opened 2021-07-30
tweak: change collateral types on STX-A and STX-B

philipdesmedt:


nieldeckx (approved):
My feedback would be:
- Use 400 as collateral-to-debt-ratio for STX-A, as this number is shown in UI
- Use 300 as collateral-to-debt-ratio for STX-B to reduce risk at launch

But that's more gut feeling than anything else so approved the PR ;)

philipdesmedt:
@nieldeckx agree - will change!

PR #160, opened 2021-08-08
tweak: use guardian address instead of DAO address to toggle shutdown

philipdesmedt:


nieldeckx (approved):
This allows us to remove the ability to turn on the emergency switches right?

philipdesmedt:
@nieldeckx correct - this moves it from the DAO keys to the guardian keys, and we will be able to burn the guardian keys in the future if we don't want to rely on the emergency switches anymore

PR #162, opened 2021-08-10
feat: make collecting fees on swap optional

philipdesmedt:


clarity/contracts/arkadiko-swap-v1-1.clar
```diff
@@ -201,7 +203,7 @@
         balance-y: u0,
         fee-balance-x: u0,
         fee-balance-y: u0,
-        fee-to-address: (contract-call? .arkadiko-dao get-payout-address),
+        fee-to-address: none, ;; TODO: (contract-call? .arkadiko-dao get-payout-address),
```
nieldeckx, line 206: TODO can be removed? Default is none, but dao owner can set it later.

nieldeckx (commented):
LGTM, just one comment.

clarity/contracts/arkadiko-swap-v1-1.clar
philipdesmedt: @nieldeckx I think we discussed with Ozzy that we should let the 5bp go to the DAO and use that to distribute stDIKO staking rewards

nieldeckx:
@philipdesmedt Code is not compiling, error on swap L219. Also not sure why you made the fee-to-address optional as it's now set when creating a pair.

philipdesmedt:
@nieldeckx yes I'll fix this branch.

The principal is optional because it allows us to switch between strategies (e.g. distribute as stDIKO, do nothing and give 25 bp to LPs or take 5bp for protocol revenue)

PR #163, opened 2021-08-10
tweak: change liquidator fees in favour of keepers

philipdesmedt:


nieldeckx (changes requested):
Auction engine tests are failing. Because total debt to raise changed.

nieldeckx:
@philipdesmedt Stacker test is failing.

PR #170, opened 2021-08-15
Extra code comments

nieldeckx:
Added comments to the following contracts:
- Stake registry
- All stake pools
- DIKO init
- DIKO guardian
- Governance
- Swap 
- Vault rewards

I assume that we only need comments for public functions which can be used by normal protocol users (non admin functions).

Couple of questions:
- Can we add any type after @post? How will this be used? ex:`;; @post auction; a new auction will be created that runs for one day`
- Should we add the full version number to the contract? ex: `;; @version 1.1` instead of `;; @version 1` on auction-engine?


philipdesmedt:
@nieldeckx great! will review & merge this today. On your questions:

- Can we add any type after @post? How will this be used? ex:;; @post auction; a new auction will be created that runs for one day
Yeah, I think so. We can basically choose this.

- Should we add the full version number to the contract? ex: ;; @version 1.1 instead of ;; @version 1 on auction-engine?
Good idea, I will change this.

philipdesmedt:
@nieldeckx can you fix merge conflict on this one?

nieldeckx:
@philipdesmedt merged master and fixed conflicts.

PR #176, opened 2021-08-17
fix: lower read length in bytes for calculation of collateralization

philipdesmedt:


clarity/contracts/arkadiko-collateral-types-v1-1.clar
```diff
@@ -112,6 +112,7 @@
   )
 )
 
+;; TODO - this is not in use
```
nieldeckx, line 115: Can we remove this method? If governance wants to update the collateral types, an updated contract can be pushed I guess? 

Or we could keep this method for now and make sure the dao-owner can call this so we can quickly update if needed.

clarity/contracts/arkadiko-freddie-v1-1.clar
```diff
@@ -119,7 +119,12 @@
 ;; @param coll-type; contract that contains the collateral types that can be used for a vault
 ;; @param oracle; the oracle implementation that provides the on-chain price
 ;; @post uint; returns the collateralization ratio
-(define-public (calculate-current-collateral-to-debt-ratio (vault-id uint) (coll-type <collateral-types-trait>) (oracle <oracle-trait>))
+(define-public (calculate-current-collateral-to-debt-ratio
+  (vault-id uint)
+  (coll-type <collateral-types-trait>)
+  (oracle <oracle-trait>)
+  (include-stability-fees bool)
```
nieldeckx, line 126: REQ: need to update method comments

nieldeckx (changes requested):
Can you explain the problem, and why the extra parameter fixes it?

philipdesmedt:
@nieldeckx yes - basically there is a read length in number of bytes (100K bytes if I recall correctly) per read-only method through RPC.

Due to the addition of extra vault values (the pox settings I think), we went over 100K bytes read in the read only method... so I needed to fix this somehow. The easiest was to not ask for the stability fee calculation from the front-end when requesting the collateralization ratio to show in the UI.. but still use it to calculate liquidations in the liquidator. Makes sense?

clarity/contracts/arkadiko-collateral-types-v1-1.clar
philipdesmedt: @nieldeckx what should we do? there's also the `change-risk-parameters` method that can be called by the DAO keys...

philipdesmedt:
@nieldeckx going to merge this - we can resolve the TODO in a later PR since it is not related.

PR #178, opened 2021-08-18
Update Vault Manage UI

Macxim:
Screen Shot 2021-08-18 at 11 24 13 AM

There is probably still room for improvement but I think it's a good first draft.

Niel said something important:

> For the left side, I think we need to make 3 scenarios very clear:
> 1) User has stacking enabled + is stacking -> can not withdraw funds, can disable stacking
> 2) User has stacking disabled + is not stacking -> can withdraw funds, can enable stacking
> 3) User has disabled stacking, but stacking still in progress -> make clear that he will be able to withdraw when stacking cycle ends


philipdesmedt:
@Macxim looks nice!

Isn't the toggle on the left bottom in the Stacking panel the same as the "Unstack ->" action?

Macxim:
@philipdesmedt Yeah, you're right. We can remove the toggle.

nieldeckx:
@Macxim 
- Fixed a small bug where STX and liquidation price were reversed.
- Noticed the collateral to debt ratio is always 0 because the code has been put in comments?
- When creating a vault without stacking enabled, the overview page still shows stacking as being enabled.
- Need to make the stacking cycle end date dynamic.

Let me know if you need any help with the above and if there are other things to do :)

philipdesmedt:
@nieldeckx the collateral to debt ratio is 0 cause it had that read bug that I fixed in the other PR

Macxim:
@nieldeckx Thanks for the fix!
I think I would need help with this: `Need to make stacking STX and cycle end date dynamic.`
I removed the Switch component.

philipdesmedt:
@Macxim @nieldeckx I can look at the `Need to make stacking STX and cycle end date dynamic.` - this might require a combination of the stacking data + some info from the stacking contract... since tokens are not stacked for 1 cycle, but 4 cycles (3 + 1 cooldown). Me or @nieldeckx can look into it.

philipdesmedt:
@Macxim now the DIKO vault rewards are gone though. Could you add them on the main Vaults page? Can be in a separate PR. I'll finish this one up right now.

Macxim:
@philipdesmedt Yeah, I didn't forget about it, I'll do it in a separate PR, we talked about it with @nieldeckx yesterday.

philipdesmedt:
@Macxim cool!

philipdesmedt:
@Macxim @nieldeckx made some changes... can you review?

Macxim:
Looks good to me for now, we can revisit later 👍

nieldeckx (approved):
Looks good!

PR #179, opened 2021-08-18
Take stDIKO balance into account when creating a proposal

nieldeckx:


philipdesmedt:
@nieldeckx I adapted the front-end and fixed some small bugs, see https://github.com/arkadiko-dao/arkadiko/pull/179/commits/2fe2e549e2bad5e4c4401c79df1c2599a5d77410. Can you check if good?

Also from the front-end you can only use DIKO now to vote I think, but we can fix that in another PR. I need to send the final smart contracts for audit.

nieldeckx:
@philipdesmedt Apologies, should have checked if parameters are correct. Thx for the update. 
Have added a new ticket for myself to update the FE so we can close this PR.

philipdesmedt:
@nieldeckx cool! no worries

I'll merge this and try to fix Maxime's PR tomorrow morning

Issue #182, opened 2021-08-19
Show DIKO/stDIKO ratio


nieldeckx:
- On stake page
- When voting with stDIKO it should be clear how many votes will be added

PR #183, opened 2021-08-19
Add ability to vote with stDIKO to UI

nieldeckx:
Functional for now. Will need some styling later.


philipdesmedt (changes requested):
Does not work for me
image

I deployed from scratch, staked 5000 DIKO and tried to vote on the proposal with these 5000 stDIKO. u3401 == not authorized in the governance contract

philipdesmedt:
@nieldeckx never mind, the start block was not reached yet so it throws an error :D gonna test with a lower start block

philipdesmedt (changes requested):
Never mind my last review, this works. Got two comments:

- When I vote with 5K stDIKO, it converts into ~5900 stDIKO votes. When the proposal ends, will I get that same amount back? Or which amount will I get back when we run the `return-votes-to-member` method?
- Can you add post conditions to the Stacks transactions in the front-end? There are some examples of this already throughout the codebase (e.g. check creating a vault or the swap for sip10).

nieldeckx:
@philipdesmedt 
- You get back back the same amount of stDIKO as you used to vote. `return-votes-to-member` has a token parameter so this method can be used to return both DIKO and/or stDIKO to the voter. This is covered by a test.
- Ok, will do.

As it's not super clear how much votes you will cast when using stDIKO, I created the following ticket to optimise:
https://github.com/arkadiko-dao/arkadiko/issues/182

nieldeckx:
@philipdesmedt double checked and the post condition is already there:
Screenshot 2021-08-23 at 20 23 25

philipdesmedt:
@nieldeckx that is not a post condition. See https://docs.stacks.co/understand-stacks/transactions#post-conditions

nieldeckx:
@philipdesmedt Oops, thx for the info. Added the post conditions.

PR #187, opened 2021-08-22
Test models

nieldeckx:


nieldeckx:
Made all tests a lot easier to read by using separate models and helper functions. Once this PR is merged I will start adding more tests, and add documentation where needed.

scripts/proposal-change-stacker.js
```diff
@@ -9,8 +9,8 @@ async function transact() {
   const list = tx.listCV([
     tx.tupleCV({
       'name': tx.stringAsciiCV("stacker"),
-      'address': tx.standardPrincipalCV("ST1PQHQKV0RJXZFY1DGX8MNSNYVE3VGZJSRTPGZGM"),
-      'qualified-name': tx.standardPrincipalCV("ST1PQHQKV0RJXZFY1DGX8MNSNYVE3VGZJSRTPGZGM", "arkadiko-mock-stacker-v1-1"),
+      'address': tx.standardPrincipalCV(deployer.address),
```
philipdesmedt, line 12: REQ: I don't think we have access to `deployer` in scripts?

philipdesmedt (changes requested):

nieldeckx:
Great catch. Fixed it.

PR #188, opened 2021-08-22
feat: onboarding

philipdesmedt:
This is work in progress.

@Macxim can continue on this branch


scripts/utils.js
```diff
@@ -38,7 +38,7 @@ async function getNonce(address) {
 
 function resolveUrl() {
   if (env === 'mocknet') {
-    return 'http://localhost:20080';
+    return 'http://localhost:3999';
```
victorcoder, line 41: back to `3999`?

philipdesmedt: "Unfortunately" yes. The clarinet release went back to 3999 which is the default. Perhaps we can configure it through an ENV variable so every contributor can choose what to do

victorcoder (approved):
LGTM

nieldeckx:
Awesome 👍 

@Macxim Copy is important here. Suggest you work with Pieter on this so it can be finished completely.

philipdesmedt:
@Macxim @nieldeckx this is ready for review imho

added a check in localStorage to redirect automatically if the user is logged in & hasn't done the onboarding yet.

philipdesmedt:
@Macxim @nieldeckx actually don't merge yet since I found a bug when you are logging in and it doesn't redirect yet

Macxim:
It's still missing a final message to _close_ the onboarding and invite the user to use the app.

philipdesmedt:
@Macxim good point - can you add that?

Macxim:
Will do in a bit.

Macxim:
Added the end step of the onboarding. Feel free to adapt the copy if needed!

philipdesmedt:
I fixed a bug where it wouldn't redirect you to the onboarding when logging in.

@Macxim last page looks great. The only thing is I have to click "Next" twice on the last governance page to get to the last page. Do you have this as well?

Macxim:
Should be fixed now 🚀

philipdesmedt:
@nieldeckx any last comments on this?

Macxim:
I would like to fix the responsive 😄 But I guess it can be done in another PR.

philipdesmedt:
@Macxim yeah - although responsiveness is not as high prio since the Hiro Web Wallet only works on FF, Chrome and Brave so far...

Macxim (approved):

philipdesmedt:
Just ran through this again. LGTM. Gonna merge

PR #190, opened 2021-08-23
Disable Claim rewards when non pending rewards

victorcoder:
The button allow to execute a contract call that won't do anything.

### DIFF
--- a/web/components/mint.tsx
+++ b/web/components/mint.tsx
@@ -27,7 +27,6 @@ import { ArchiveIcon } from '@heroicons/react/outline';
 import { PlaceHolder } from './placeholder';
 import { InformationCircleIcon } from '@heroicons/react/solid';
 import { Tooltip } from '@blockstack/ui';
-import { classNames } from '@common/class-names';
 
 export const Mint = () => {
   const address = useSTXAddress();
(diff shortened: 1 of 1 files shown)


Macxim (changes requested):
Good catch!
Could you maybe just hide the button if there are no rewards?

Screen Shot 2021-08-23 at 9 08 51 PM

victorcoder:
Feedback applied

web/components/mint.tsx
```diff
@@ -388,15 +387,16 @@ export const Mint = () => {
...
-              <button 
-                type="button" 
-                className={classNames(pendingVaultRewards === 0 && 'hidden',
-                  'bg-indigo-600 hover:bg-indigo-700 inline-flex items-center px-3 py-2 ml-4 text-sm font-medium leading-4 text-white border border-transparent rounded-md shadow-sm focus:outline-none focus:ring-2 focus:ring-offset-2 focus:ring-indigo-500')}
-                onClick={() => claimPendingRewards()}
-                disabled={pendingVaultRewards === 0}
-              >
-                Claim rewards
-              </button>
+              { pendingVaultRewards > 0 &&
```
Macxim, line 390: Extra space before `{` here.
Also, we used the ternary `{condition ? () : null}` throughout the code base, maybe we should stick to it? I'm no expert, just aiming for consistency here.

philipdesmedt: I guess we don't need an `else` clause, probably remove the `null`s in other places

victorcoder: Agree with @philipdesmedt , this construct is better that ternaries with `null`, and to achieve consistency it's better to refactor the other places.

Macxim: Can you explain how/why it's better? https://kentcdodds.com/blog/use-ternaries-rather-than-and-and-in-jsx

victorcoder: JS wonderfulness 😂 , that's a compelling reason to not use the `&&` then. I'll apply the changes.

victorcoder:
Feedback applied, let me know if you want me to squash commits or you are going to to Squash and merge the PR.

philipdesmedt:
@victorcoder thanks mate! We just squash and merge here so no need :)

Macxim:
@victorcoder Good job! Thanks!

PR #191, opened 2021-08-24
Fix Withdraw Max amount Input

victorcoder:
It was pointing to an incorrect function and missing its own function.


nieldeckx:
Looks good to me. @philipdesmedt can you double check and merge if approved?

philipdesmedt:
@nieldeckx will do!

philipdesmedt:
great work, thx @victorcoder!

PR #194, opened 2021-08-26
Extra governance tests

nieldeckx:
- Prevent proposal from starting in the past
- Should not be able to return votes multiple times
- Extra tests


philipdesmedt:
@nieldeckx well done on fixing these bugs! great finds

PR #197, opened 2021-09-03
Fix: Autostack not being reflected in transaction

victorcoder:
The checked/unchecked status was not correctly being stored in the state and in consequence not being passed to the transaction.


Macxim (approved):

philipdesmedt:
@victorcoder mate you've been killing it, great bugfix, thanks!

PR #205, opened 2021-09-14
feat: add xBTC as collateral type

philipdesmedt:


philipdesmedt:
@nieldeckx @Macxim can you review this? should be ready, except for a bug in the post conditions so I use allow mode for xBTC right now. The Hiro team is aware of this bug and are fixing it asap.

I spoke to Tokensoft (xBTC team) and they can provide xBTC on mainnet within the next few weeks. I can elaborate more in a private meeting, but should be ready for our mainnet launch as a collateral type.

Macxim:
That's awesome. I'll review ASAP.

philipdesmedt:
Cool - you can get xBTC by putting your local test address here btw https://github.com/arkadiko-dao/arkadiko/pull/205/files#diff-38e1806b08d16ec258f20ced8d67d28bada5b101bfccc4534b6ef348be1a3f64R213

nieldeckx (changes requested):
This is so great! :D

When using the 'max' button on amount to collateralize there is an error "You cannot collateralize more than your balance"

Suggest to add a new row to the collateral type table that shows if loans can repay themselves. Our USP is self repaying loans so should be clear this won't be able with xBTC.

Macxim:
Just a reminder to filter out xBTC from the Swap, as we'll only use it as collateral type for now.
 
https://user-images.githubusercontent.com/1909957/133856595-2fb0b1fc-7451-4640-91e1-25c3d8128adf.mov

philipdesmedt:
@Macxim thanks, this is a bug. I'll add logic to filter it out in this PR

philipdesmedt:
@nieldeckx @Macxim fixed the bugs you found and some other ones on the decimals. Bitcoin has 8 decimals, STX and other FT usually only 6, so calculations changed.

Need to go over everything again to check.

nieldeckx:
Added extra tests for the new collateral type. Mainly focussed on the sip10 reserve as it will will now be used. 

Fixed a vulnerability in the "close-vault" function in freddie where the parameters were not checked correctly.

philipdesmedt:
This should be almost ready on the Tokensoft side. I will merge this one when they are on mainnet.

philipdesmedt:
Also need to test if liquidations & auctions work with xBTC

philipdesmedt:
@nieldeckx @Macxim I think this finally works completely with all latest updates 👀 

Need to test the auctions a bit more

philipdesmedt:
@nieldeckx for the second TODO, the way this is calculated is as follows (IDK if this is the best way):

1. Price of xBTC on Oracle is at $20,000
2. On the $20K oracle price, we discount 7% (= $18,600)
3. One lot: $1000. 1000/18600 == 0.0537634408602 ~= 0.0537

Makes sense? Or not? I'm not sure if we should do the discount on the unit (xBTC, STX, ...) or on the dollar value of the unit.. seems quite a big difference?

PR #206, opened 2021-09-15
Extra burn method for DIKO

nieldeckx:
Also made it easy to set up the stacks API port via .env files. Need to add LOCAL_STACKS_API_PORT to .env files in both web and scripts folder.


philipdesmedt (approved):
LGTM

PR #215, opened 2021-09-19
fix: LP tokens wSTX-DIKO and wSTX-USDA being staked by users are lost [0F-ARK-006]

philipdesmedt:
See https://github.com/0f/0f-ark-2108/issues/7


nieldeckx (approved):
Ok. 
(Have asked auditor for more info on this, because in our tests it was working.)

philipdesmedt:
@nieldeckx Ludo has replied - tests were for the other pair that was fixed already :)

Should we add some additional tests?

nieldeckx:
Damn, stupid mistake. Will add tests for all pools.

nieldeckx:
@philipdesmedt added more tests and fixed an issue in the emergency-withdraw method

philipdesmedt:
@nieldeckx good catch 😬

PR #220, opened 2021-09-23
fix: unstake DIKO popup shows wrong max amount

nieldeckx:
@Macxim fix for the bug you discovered


Macxim:
We probably also want to change the details of the input, right? @nieldeckx 
See above: https://github.com/arkadiko-dao/arkadiko/pull/220/commits/1438a5d263fb03dc9ebc90e45a05e64e1ddfe10b

nieldeckx:
@Macxim Yes, correct. Great catch!

Macxim (approved):
LFG 🚀

PR #222, opened 2021-09-24
fix: set correct parameters for post conditions on add liquidity

philipdesmedt:


philipdesmedt:
going to merge to test on testnet

PR #225, opened 2021-09-26
fix: improve post conditions on swap and auction bidding

philipdesmedt:


web/components/add-swap-liquidity.tsx
```diff
@@ -249,20 +247,6 @@ export const AddSwapLiquidity: React.FC = ({ match }) => {
         )
       );
     }
-    // TODO: do we need this PC?
```
nieldeckx, line 252: Nice to have I guess. Not critical.

nieldeckx (approved):
Great!

PR #226, opened 2021-09-27
fix: defective funds migration on contract upgrade [0f-ARK-007]

philipdesmedt:
see issue https://github.com/0f/0f-ark-2108/issues/8


nieldeckx (approved):
Perfect

PR #227, opened 2021-09-28
fix: oracle price decimals [0f-ARK-011]

nieldeckx:


nieldeckx:
@philipdesmedt oracle prices now have 6 decimals. Not that many changes in code (mostly comments and tests changed).

Is it correct that in the following function the collateral-amount set does not really matter but just needs to be larger than the debt?
Screenshot 2021-09-28 at 14 29 23

philipdesmedt:
@nieldeckx that is a good catch lol.. the debt auction is done to sell off DIKO tokens. We need to fetch the price of DIKO from the oracle (or straight from our swap) there since now it is hard-coded to 10 cents.

clarity/contracts/arkadiko-auction-engine-v1-1.clar
```diff
@@ -244,14 +244,14 @@
...
 ;; @param auction-id; the ID of the auction in which the collateral will be sold
-(define-read-only (discounted-auction-price (price-in-cents uint) (auction-id uint))
+(define-read-only (discounted-auction-price (price uint) (auction-id uint))
   (let (
     (auction (get-auction-by-id auction-id))
-    (discount (* price-in-cents (get discount auction)))
+    (discount (* price (get discount auction)))
   )
-    (ok (/ (- (* u100 price-in-cents) discount) u100))
+    (ok (/ (- (* u100 price) discount) u1000000))
```
philipdesmedt, line 254: Q: we divide by one million here (6 decimals)

what if we handle bitcoin, which has 100M denomination (1 sat = 8 decimals)?

clarity/contracts/arkadiko-sip10-reserve-v1-1.clar
```diff
@@ -36,9 +36,9 @@
   (ucollateral uint)
   (oracle <oracle-trait>)
 )
-  (let ((price-in-cents (unwrap-panic (contract-call? oracle fetch-price token))))
+  (let ((price (unwrap-panic (contract-call? oracle fetch-price token))))
     (if (> debt u0)
-      (ok (/ (* ucollateral (get last-price-in-cents price-in-cents)) debt))
+      (ok (/ (/ (* ucollateral (get last-price price)) debt) u10000))
```
philipdesmedt, line 41: Q: where does the division by 10K come from?

philipdesmedt (changes requested):

clarity/contracts/arkadiko-auction-engine-v1-1.clar
nieldeckx: The oracle now has 6 decimals to save a price. So BTC price in $ will have 6 decimals, and does not depend on the amount of decimals of the token itself.

philipdesmedt: mm ok.. so which price do we push to the oracle if bitcoin is $50,000 dollar?

nieldeckx: In that case we set $50.000 with 6 decimals so 50000000000.

( Previously you had the price in cents, so 2 decimals )

philipdesmedt: OK got it! that makes sense

philipdesmedt:
@nieldeckx let me know when this is ready for review and I'll go through thoroughly (also merge conflict needs fixing)

nieldeckx:
@philipdesmedt  Should be good to review.

- When redeeming a bad debt lot, we will only mint just enough tokens for that lot. 

- The collateral-amount for a bad debt auction does not really matter because when bidding `get-minimum-collateral-amount` will make sure the collateral amount is correctly calculated. So we just need to make sure that the collateral amount on the bad debt auction is definitely enough which is why I set a hard-coded price of $1M per DIKO.

- Both the bidding and redeem function can close a vault. But when closing it could be that there is no debt to burn left which will cause the call to fail. Added a check to avoid burning 0 tokens. But wondering why we want to close a vault in the redeem function? When removing the close-vault call in the redeem function all tests still work. Any idea?

The 3 changes above should be covered in the test.

philipdesmedt:
@nieldeckx great, thanks, will review!

On your last comment, I think I added a call to `close-vault` in the redeem function to optimise things. When someone redeems a lot and we see that the auction can be closed, we do that, so other people or we don't have to take care of this anymore. It's saving an extra on-chain TX. I could be wrong here and maybe it results in bugs tho...

nieldeckx:
@philipdesmedt yeah suggest we keep the `close-vault` call. The extra check prevents it from failing if amount is 0 so all good.

philipdesmedt (approved):
No comments. Great work!

philipdesmedt:
Don't merge this yet though, I need to test the front-end. Don't think it works yet.

philipdesmedt:
I think this works now. Will merge later

philipdesmedt:
Darn - fixed a few more bugs, both in the oracle changes for the front-end as well as some auction bugs :)

Need to go over `master` again soon once everything is merged.

PR #239, opened 2021-09-29
Improve list limits [0f-ARK-010]

nieldeckx:
Changed:
- vault-entries: 500 -> 20

Others:
- winning-lots: 100
- auction-ids: 1500
- pairs-list: 2000
- stacking-payout-lots: 500
- proposal-ids: 100
- contract-changes: 10

I think they are fine for now. We will need to find a way to have more auctions and proposals but this would require too much refactoring at this point.

### DIFF
--- a/clarity/contracts/arkadiko-vault-data-v1-1.clar
+++ b/clarity/contracts/arkadiko-vault-data-v1-1.clar
@@ -177,7 +177,7 @@
       }
     )
     (if (is-none (index-of entries lot-index))
-      (map-set stacking-payout-lots { vault-id: vault-id } { ids: (unwrap-panic (as-max-len? (append entries lot-index) u20)) })
+      (map-set stacking-payout-lots { vault-id: vault-id } { ids: (unwrap-panic (as-max-len? (append entries lot-index) u500)) })
       true
     )
     (ok true)
--- a/clarity/contracts/arkadiko-vault-data-v1-1.clar
+++ b/clarity/contracts/arkadiko-vault-data-v1-1.clar
@@ -25,7 +25,7 @@
   auction-ended: bool,
   leftover-collateral: uint
 })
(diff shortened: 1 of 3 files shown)


clarity/contracts/arkadiko-vault-data-v1-1.clar
```diff
@@ -177,7 +177,7 @@
       }
     )
     (if (is-none (index-of entries lot-index))
-      (map-set stacking-payout-lots { vault-id: vault-id } { ids: (unwrap-panic (as-max-len? (append entries lot-index) u500)) })
+      (map-set stacking-payout-lots { vault-id: vault-id } { ids: (unwrap-panic (as-max-len? (append entries lot-index) u20)) })
```
philipdesmedt, line 180: REQ: this should stay 500, no?

philipdesmedt (changes requested):

philipdesmedt:
@nieldeckx thanks for going over this!

Do you think 20 vaults per principal isn't a little low? We might have to block vault 21 in the UI since I'm sure there will be apes making more than 20 vaults :D

clarity/contracts/arkadiko-vault-data-v1-1.clar
nieldeckx: Yes indeed. Thx.

nieldeckx:
> @nieldeckx thanks for going over this!
> 
> Do you think 20 vaults per principal isn't a little low? We might have to block vault 21 in the UI since I'm sure there will be apes making more than 20 vaults :D

I could not think of a situation where I would create over 20 vaults. Can you?

philipdesmedt:
@nieldeckx rationally I can't. People aren't rational though lmao. We can keep it to 20 but perhaps we should be ready to block 21 and onwards from the UI?

philipdesmedt (approved):

PR #244, opened 2021-09-30
Unnecessary statements [0f-ARK-013]

nieldeckx:


clarity/contracts/arkadiko-auction-engine-v1-1.clar
```diff
@@ -671,7 +671,6 @@
     (map-set auctions
       { id: auction-id }
       (merge auction {
-        total-debt-burned: (get total-debt-raised auction),
```
philipdesmedt, line 674: Q: can you elaborate why we can remove this?

philipdesmedt (changes requested):

clarity/contracts/arkadiko-auction-engine-v1-1.clar
nieldeckx: The `total-debt-burned` is updated in the `close-auction` function, which is the only place from which this `extend-auction` function is called.

I have added this line again so we can close this PR. Will create a new PR with additional auction tests.

PR #249, opened 2021-10-01
feat: Add info about STX stacking in Vault creation flow

Macxim:
Screen Shot 2021-10-01 at 12 28 49 PM

### DIFF
--- a/web/components/create-vault-confirm.tsx
+++ b/web/components/create-vault-confirm.tsx
@@ -1,5 +1,5 @@
 import React from 'react';
-import { QuestionMarkCircleIcon, ExternalLinkIcon } from '@heroicons/react/solid';
+import { QuestionMarkCircleIcon, ExternalLinkIcon, ExclamationIcon } from '@heroicons/react/solid';
 
 
 export const CreateVaultConfirm = ({ setStep, coinAmounts, setCoinAmounts }) => {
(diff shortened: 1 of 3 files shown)


philipdesmedt (approved):
Good for me.

Macxim:
@philipdesmedt As discussed, we still need to add an approximate date of end of cycle to let the users know when coins are unlocked. @nieldeckx Would you mind?

philipdesmedt:
@Macxim this should be pretty easy, we already fetch the PoX dates from stacking.club API and put that in the global React state... so you can access that with `{state.endDate}` + 2 weeks (`const [state, setState] = useContext(AppContext);` at the top).

Macxim:
@philipdesmedt Done

nieldeckx:
The `{state.endDate}` indicates when the current stacking cycle will end.  So currently it shows Oct 6th.

@philipdesmedt I thought we would be stacking for 3 cycles? In that case the calculation would be something like `{state.endDate} + 6 weeks stacking + 2 week cooldown`?

philipdesmedt:
@nieldeckx yes, correct... but we have to take into account in which cycle we stacked their STX, so it can be anything between 1 and 6 weeks

nieldeckx:
@Macxim just discussed with Philip. 

When creating a vault and stacking STX:
- actual stacking starts in next cycle (so at {state.endDate})
- we stack for 3 cycles where each cycle is 2 weeks (so 6 weeks total)
- there is a cooldown of 2 weeks

So when creating a vault we can show that the funds will be locked and available again at:
`{state.endDate} + 6 weeks stacking + 2 week cooldown`

Macxim:
What do you think?

Screen Shot 2021-10-06 at 9 04 05 PM

philipdesmedt:
@Macxim can't you show the exact end date?

Macxim:
Done @philipdesmedt 
Screen Shot 2021-10-07 at 8 58 07 AM

PR #251, opened 2021-10-02
fix: allow setting a custom block height for DIKO vault rewards

philipdesmedt:


nieldeckx (approved):
Great. We can do the same in the staking pools.

philipdesmedt:
@nieldeckx all right, thanks for approval! I will do this for the staking pools in another PR

PR #253, opened 2021-10-04
fix: allow specifying block height in the future for when staking starts

philipdesmedt:


nieldeckx (approved):
@philipdesmedt should we add a TODO to set the "last-reward-increase-block" or have you added it to the list for launch?

philipdesmedt:
@nieldeckx I have it in the list, but gonna add it here too, thx

PR #254, opened 2021-10-04
Split Stake section into two parts

Macxim:


nieldeckx:
The improvements are great. 👍  Also like how there's more focus on DIKO staking now.

Additionally it might be a good idea to mention that the cooldown period is 10 days in the tooltip.

In the stake popup we could explain to the user that staked DIKO will be stuck for at least 10 days.
Screenshot 2021-10-05 at 13 25 12

Macxim:
Good remarks, I'll add them, thanks.

Macxim:
@nieldeckx Done.

### Modal
Screen Shot 2021-10-05 at 4 19 56 PM

### Tooltip
Screen Shot 2021-10-05 at 4 20 20 PM

nieldeckx:
@Macxim cool! 
Sorry, maybe small change is to use "locked" instead of "stuck"?

nieldeckx (approved):
Nice!

PR #263, opened 2021-10-11
Fix react warnings for input checkboxes in vault creation

Macxim:
Screen Shot 2021-10-11 at 9 13 02 PM

@victorcoder, could you have a look as well? Is it the best solution to fix those warnings?


victorcoder:
@Macxim in my case I'm not getting the warnings, your changes in the PR implements what the error message is complaining about. Is this correct or are you still getting the warnings?

PR #272, opened 2021-10-13
feat: add disclaimer in onboarding

philiphacks:
image

Feel free to adapt/change.

We should add a disclaimer on the marketing website as well.


philipdesmedt:
@nieldeckx @Macxim thoughts?

Macxim:
@philipdesmedt thoughts?

Screen Shot 2021-10-14 at 4 18 05 PM

PR #310, opened 2021-10-25
feat: add API to show TVL & prices of Arkadiko Pools

philipdesmedt:
Added a simple Rails API in the same repo.

Will have some simple models in an open API that can be queried. Cron jobs will run every 10 minutes (sync per Stacks anchor block) to query the TVL + prices


philipdesmedt:
@Macxim the airbnb package looks nice! will have a look

PR #316, opened 2021-10-28
fix: swap v2

philipdesmedt:
Hi everyone!

For those of you coming here through an on-chain governance vote, welcome :)

This PR contains the code changes for AIP-1-swap-v2. It replaces the defect `arkadiko-swap-v1-1.clar` contract with a `arkadiko-swap-v2-1.clar` contract in 10 days (block height 37281). In the governance vote, we will change the following `(contract-changes (list (tuple (address SP2C2YFP12AJZB4MABJBAJ55XECVS7E4PMMZ89YZR) (can-burn true) (can-mint true) (name "swap") (qualified-name SP2C2YFP12AJZB4MABJBAJ55XECVS7E4PMMZ89YZR.arkadiko-swap-v2-1)))`

The following was changed:
- Added an emergency shutdown toggle that disables all critical swap functions (adding liquidity, removing liquidity, swapping)
- An LP token can only be used once per pair
- Made the swap permissioned so only pairs can be added through the DAO after vetting the SIP10 tokens
- Added a toggle to disable/enable trading on a pair
- Refactored the wSTX comparison to not use a string but compare the contract addresses

All changes are reflected in `arkadiko-swap-v1-1.clar`. We will create and deploy `arkadiko-swap-v2-1.clar` once this is production-ready.

If you want to join the discussion, you can comment on technical details here or join the discussion on Discord: https://discord.gg/JbbRWDwRbU


clarity/contracts/arkadiko-swap-v1-1.clar
```diff
@@ -249,10 +290,18 @@
       )
       pair-already-exists-err
     )
```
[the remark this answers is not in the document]
philipdesmedt: Agree - we are going to make the adding of pairs to the swap permissioned.

clarity/contracts/arkadiko-swap-v1-1.clar
```diff
@@ -249,10 +290,18 @@
...
+    (asserts!
+      (and
+        (is-eq (unwrap-panic (contract-call? .arkadiko-dao get-emergency-shutdown-activated)) false)
+        (is-eq (var-get swap-shutdown-activated) false)
+      )
+      (err ERR-EMERGENCY-SHUTDOWN-ACTIVATED)
+    )
+    (asserts! (is-eq (var-get can-add-pairs) true) (err ERR-NOT-AUTHORIZED))
```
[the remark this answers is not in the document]
philipdesmedt: Yes - I am going to refactor this slightly to allow adding pairs only through governance.

psq:
it is **IMPERATIVE** to check the `token-x` and `token-y` traits in **ALL*** functions, otherwise `swap`, `add-to-position`, and `reduce-position` can be attacked in similar fashion with fake tokens, even with the check for the lp token trait.

psq:
Also, in order to migrate the pairs from v1, you are going to need a special, permissioned, function to create a pair with the right amount of lp tokens, but without minting these tokens (the balances/total-supply in the lp tokens are correct, even after the hack.

clarity/contracts/arkadiko-swap-v1-1.clar
```diff
@@ -505,3 +597,15 @@
     (ok (list fee-x fee-y))
   )
 )
+
+;; temporary method to allow an attack on a malicious LP minter, if any
+(define-public (attack-and-burn (swap-token-trait <swap-token>) (address principal) (amount uint))
```
psq, line 602: this may be going too far, especially if you burn the lp token like this, you will not be able to retrieve the tokens (if they are real) for that position
instead, consider just moving the lp tokens (burn + mint) to a safe address

clarity/contracts/arkadiko-swap-v1-1.clar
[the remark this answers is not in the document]
psq: this is not very "permissionless", but the contracts used for lp contracts should be contracts that have been vetted  (maybe via a pre-registration, and the meta data can be set once before someone creates the pair

philipdesmedt:
@psq we will be vetting all new tokens (and thus pairs) through the DAO, so moving to slightly more permissioned model for adding pairs. This way we can trust the `swap`, `add-to-position` and `reduce-position` methods once a pair is added.

The migration method is a good point - I will add that.

clarity/contracts/arkadiko-swap-v1-1.clar
philipdesmedt: Talked about this with psq in DM and is practically infeasible.

clarity/contracts/arkadiko-swap-v1-1.clar
```diff
@@ -128,6 +145,17 @@
   )
 )
 
+(define-read-only (is-registered-swap-token (swap-token principal))
+  (is-some (map-get? registered-swap-tokens { swap-token: swap-token }))
+)
+
+(define-private (register-swap-token (swap-token principal))
+  (begin
+    (asserts! (is-eq (is-registered-swap-token swap-token) false) (err ERR-WRONG-SWAP-TOKEN))
```
LNow, line 154:
```suggestion
    (asserts! (not (is-registered-swap-token swap-token)) (err ERR-WRONG-SWAP-TOKEN))
```
`(not x)` is cheaper to execute than `(is-eq x false)` and I also thing it is more readable.

clarity/contracts/arkadiko-swap-v1-1.clar
```diff
@@ -166,15 +194,22 @@
     )
     (asserts! (and (> x u0) (> new-y u0)) (err ERR-INVALID-LIQUIDITY))
     (asserts! (is-eq swap-token (contract-of swap-token-trait)) (err ERR-WRONG-SWAP-TOKEN))
+    (asserts!
+      (and
+        (is-eq (unwrap-panic (contract-call? .arkadiko-dao get-emergency-shutdown-activated)) false)
+        (is-eq (var-get swap-shutdown-activated) false)
+      )
```
LNow, line 201:
```suggestion
      (and
        (not (unwrap-panic (contract-call? .arkadiko-dao get-emergency-shutdown-activated)))
        (not (var-get swap-shutdown-activated))
      )
```
`(not x)` is cheaper to execute than `(is-eq x false)` and I also think it is more readable.

clarity/contracts/arkadiko-swap-v1-1.clar
```diff
@@ -242,17 +355,25 @@
...
         (is-none (map-get? pairs-data-map { token-x: token-x, token-y: token-y }))
         (is-none (map-get? pairs-data-map { token-x: token-y, token-y: token-x }))
       )
       pair-already-exists-err
     )
+    (asserts!
+      (and
+        (is-eq (unwrap-panic (contract-call? .arkadiko-dao get-emergency-shutdown-activated)) false)
+        (is-eq (var-get swap-shutdown-activated) false)
+      )
```
LNow, line 370: Extract this code into separate `read-only` function.
That way code will be a bit more readable and you'll have part of #320 already in place ;)

clarity/contracts/arkadiko-swap-v1-1.clar
```diff
@@ -311,15 +446,24 @@
...
 
     (asserts! (<= percent u100) (err u5))
+    (asserts!
+      (and
+        (is-eq (unwrap-panic (contract-call? .arkadiko-dao get-emergency-shutdown-activated)) false)
+        (is-eq (var-get swap-shutdown-activated) false)
+      )
+      (err ERR-EMERGENCY-SHUTDOWN-ACTIVATED)
+    )
+    (asserts! (is-eq (get enabled pair) true) (err ERR-PAIR-DISABLED))
```
LNow, line 456:
```suggestion
    (asserts! (get enabled pair) (err ERR-PAIR-DISABLED))
    (asserts!
      (and
        (is-eq (unwrap-panic (contract-call? .arkadiko-dao get-emergency-shutdown-activated)) false)
        (is-eq (var-get swap-shutdown-activated) false)
      )
      (err ERR-EMERGENCY-SHUTDOWN-ACTIVATED)
    )
```
Verifying pair is enabled should be done before verifying emergency shutdown. That way if pair is disabled TX will consume less execution costs and as a result will leave more room for other transactions.

Also checking if `bool` value is equal `true` is a bit excessive.

boohyunsik:
Is it really neccessary that pairs can be added through DAO? I think it is very important that anyone can add their token pair in DEX.

philipdesmedt:
@boohyunsik unfortunately anyone can implement a malicious SIP10 fungible token in Stacks that tries to transfer stuff so I think the safest option is to go to a more permissioned model for now, where tokens are vetted through a DAO-like model before they are added.

boohyunsik:
> @boohyunsik unfortunately anyone can implement a malicious SIP10 fungible token in Stacks that tries to transfer stuff so I think the safest option is to go to a more permissioned model for now, where tokens are vetted through a DAO-like model before they are added.

Thank you for fast reply! But I think every single SIP-10 token is SIP-10, like in Ehtereum, meems token is erc20. I agree that permissioned-model is more safety, but how important about freely create a new pool? Or is there other issues about SIP-10?

philipdesmedt:
@boohyunsik SIP-10 is similar to ERC20 on Ethereum yes, but there are other intricacies and it has another security model. I can elaborate on a call how the swap would be vulnerable in a permissionless model cause here it would be too much typing :)

boohyunsik:
@philipdesmedt Thank you very much! I am wondering how pool can be attacked in pemissionless model. :D

clarity/contracts/arkadiko-swap-v1-1.clar
```diff
@@ -24,6 +26,16 @@
 (define-constant no-fee-x-err (err u75))
 (define-constant no-fee-y-err (err u76))
 
+(define-data-var swap-shutdown-activated bool false)
+
+(define-public (toggle-swap-shutdown)
+  (begin
+    (asserts! (is-eq tx-sender (contract-call? .arkadiko-dao get-guardian-address)) (err ERR-NOT-AUTHORIZED))
```
LNow, line 33: `tx-sender` should be replaced with `contract-caller` or you should add to this function 1 micro STX transfer force `post-conditions` usage.

https://app.sigle.io/friedger.id/HuOT9tNQC8fTXOsK28D7e

clarity/contracts/arkadiko-swap-v1-1.clar
```diff
@@ -277,6 +388,20 @@
...
 )
 
+(define-public (toggle-pair-enabled (token-x-trait <ft-trait>) (token-y-trait <ft-trait>))
+  (let (
+    (token-x (contract-of token-x-trait))
+    (token-y (contract-of token-y-trait))
+    (pair (unwrap-panic (map-get? pairs-data-map { token-x: token-x, token-y: token-y })))
+    (pair-data { enabled: (not (get enabled pair)) })
+  )
+    (asserts! (is-eq tx-sender (contract-call? .arkadiko-dao get-guardian-address)) (err ERR-NOT-AUTHORIZED))
```
LNow, line 398: Replace `tx-sender` with `contract-caller` or add 1 micro STX transfer.

clarity/contracts/arkadiko-swap-v1-1.clar
```diff
@@ -24,6 +26,16 @@
 (define-constant no-fee-x-err (err u75))
 (define-constant no-fee-y-err (err u76))
 
+(define-data-var swap-shutdown-activated bool false)
+
+(define-public (toggle-swap-shutdown)
```
lgalabru, line 31: Should we ensure test coverage there? I was unable to find some tests using this feature.

clarity/contracts/arkadiko-swap-v1-1.clar
```diff
@@ -128,6 +137,17 @@
   )
 )
 
+(define-read-only (is-registered-swap-token (swap-token principal))
+  (is-some (map-get? registered-swap-tokens { swap-token: swap-token }))
```
lgalabru, line 141: Should we ensure that `registered` is set to `true`?

clarity/contracts/arkadiko-swap-v1-1.clar
```diff
@@ -49,8 +62,19 @@
...
+)
+
 (define-data-var pair-count uint u0)
-(define-data-var pairs-list (list 2000 uint) (list ))
+
+(define-read-only (is-shutdown-activated)
+  (and
+     (not (unwrap-panic (contract-call? .arkadiko-dao get-emergency-shutdown-activated)))
+     (not (var-get swap-shutdown-activated))
+  )
```
lgalabru, line 76: The boolean logic in this function is a bit counter intuitive / name of the function is misleading.
The result or your operation is indicating that shutdown is **not** activated.

clarity/contracts/arkadiko-swap-v1-1.clar
```diff
@@ -211,8 +232,88 @@
   (ok (var-get pair-count))
 )
 
-(define-read-only (get-pairs)
-  (ok (map get-pair-contracts (var-get pairs-list)))
+(define-public (migrate-create-pair (token-x-trait <ft-trait>) (token-y-trait <ft-trait>) (swap-token-trait <swap-token>) (pair-name (string-ascii 32)) (shares-total uint))
```
lgalabru, line 235: Not sure to understand why this function is being introduced, could you comment on this? What would have been the issue with using the `create-pair` method?

clarity/contracts/arkadiko-swap-v1-1.clar
```diff
@@ -9,6 +9,8 @@
 (define-constant ERR-INVALID-LIQUIDITY u202)
```
lgalabru, line 9: Should we update the name to v2, and make sure that all the other contracts are using the right version?
Looking at the code, the method `payoff-vault-debt` in `arkadiko-stacker-payer-v1-1` is has a hard coded reference to `arkadiko-swap-v1-1`, there could be a cascade there.

clarity/contracts/arkadiko-swap-v1-1.clar
```diff
@@ -211,8 +232,88 @@
...
+    (asserts! (is-ok (contract-call? token-x-trait transfer x tx-sender contract-address none)) transfer-x-failed-err)
+    (asserts! (is-ok (contract-call? token-y-trait transfer y tx-sender contract-address none)) transfer-y-failed-err)
+
+    (map-set pairs-data-map
+      { token-x: token-x, token-y: token-y }
+      (merge pair-data { balance-x: new-balance-x, balance-y: new-balance-y })
+    )
+
+    (ok true)
+  )
```
lgalabru, line 315: Is there a way to design the migration differently with a free `begin` statement, executed once when the contract is being deployed, that would fetch the data from the swap v1?
I don't feel particularly good with the idea of adding more public function (I can see that these are guarded tho, but still).

clarity/contracts/arkadiko-swap-v1-1.clar
```diff
@@ -505,3 +651,15 @@
     (ok (list fee-x fee-y))
   )
 )
+
+;; temporary method to allow an attack on a malicious LP minter, if any
+(define-public (attack-and-burn (swap-token-trait <swap-token>) (address principal) (amount uint))
```
lgalabru, line 656: Is this function considered as production code?

clarity/contracts/arkadiko-swap-v1-1.clar
```diff
@@ -277,6 +326,18 @@
   )
 )
 
+(define-public (toggle-pair-enabled (token-x-trait <ft-trait>) (token-y-trait <ft-trait>))
```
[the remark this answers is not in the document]
lgalabru: Should we ensure test coverage there? I was unable to find some tests using this feature.

clarity/contracts/arkadiko-swap-v1-1.clar
philipdesmedt: We have tests on that, see here: https://github.com/arkadiko-dao/arkadiko/pull/316/files#diff-0e4d987d6791a7d5251a79a65aec837f12eb70f300be849696300b2564f5d464R753

clarity/contracts/arkadiko-swap-v1-1.clar
philipdesmedt: I'm don't think it matters since you can't unregister a pair (there is no `map-delete` anywhere)

philipdesmedt: As a result, I think `(is-some (map-get? registered-swap-tokens { swap-token: swap-token }))` is enough as a check, since if it exists in the map, it is registered and it can't be unregistered

clarity/contracts/arkadiko-swap-v1-1.clar
philipdesmedt: Agree - I have renamed the method.

clarity/contracts/arkadiko-swap-v1-1.clar
philipdesmedt: We needed a migration method so we could set the `total-shares` in the pair. The `create-pair` initialises the pair and sets the total shares to zero. This allows us to create a pair, set the total shares, and then later add liquidity through it with `migrate-add-liquidity`

We preferred this code duplication over changing the arguments in the `create-pair` method since `create-pair` will be used often in the future.

clarity/contracts/arkadiko-swap-v1-1.clar
philipdesmedt: Yes - in fact we are copying all this code into a `swap-v2.clar` file once it's ready for deployment, and then we'll be retiring the `arkadiko-stacker-payer-v1-1` in favor of a contract where the users claim the yield themselves instead of us putting the yield straight into the vault (since it was not scalable). SEe here https://github.com/arkadiko-dao/arkadiko/pull/319

clarity/contracts/arkadiko-swap-v1-1.clar
philipdesmedt: @nieldeckx thoughts?

I think fetching from the v1 contract would be hard tho.. all the funds are also on a contract we own ATM.

clarity/contracts/arkadiko-swap-v1-1.clar
philipdesmedt: Yes - but it's supposed to be only executable up to a certain block height (see the TODO). Any thoughts on this?

clarity/contracts/arkadiko-swap-v1-1.clar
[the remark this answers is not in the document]
philipdesmedt: We have tests on this, see https://github.com/arkadiko-dao/arkadiko/pull/316/files#diff-0e4d987d6791a7d5251a79a65aec837f12eb70f300be849696300b2564f5d464R729

clarity/contracts/arkadiko-swap-v1-1.clar
philipdesmedt: Alternatively we could also allow doing this only up until a certain block height.

philipdesmedt:
This is deployed
https://explorer.stacks.co/txid/0x1e3e779fcedc0f0deac5272876bbde2c113c8626edf57606f5716e1e3f082a66?chain=mainnet

Will merge.

PR #319, opened 2021-10-29
feat: add contract for users to claim yields from PoX

philipdesmedt:
We have a contract called `arkadiko-stacker-payer-v1-1.clar` that takes care of depositing earned yield into vaults or paying off USDA debt through the AMM.

This stacker payer contract however is too heavy to work. It does not scale with the number of vaults, since it requires an on-chain transaction for each vault. Bundling the transactions is not feasible since the costs of running 1 such transaction are already significant.

Therefore, we built this new yield claimer. Each vault owner can claim the yield for their vault through this simple contract.


nieldeckx:
@Macxim vault page will now show the yield and buttons to claim it (either add to collateral or use to pay back debt). Can you please fix the alignment of the buttons? Feel free to restructure info as well.

philipdesmedt:
@nieldeckx what do you think of automatically stacking the new collateral + old collateral when stacking is on? I think 99% would want to compound.

clarity/contracts/arkadiko-claim-yield-v1-1.clar
```diff
@@ -0,0 +1,178 @@
...
+  )
+    (asserts! (is-eq tx-sender (get owner vault)) (err ERR-NOT-AUTHORIZED))
+    (asserts! (is-eq (contract-of reserve) (unwrap-panic (contract-call? .arkadiko-dao get-qualified-name-by-name "stx-reserve"))) (err ERR-NOT-AUTHORIZED))
+    (asserts! (is-eq (contract-of coll-type) (unwrap-panic (contract-call? .arkadiko-dao get-qualified-name-by-name "collateral-types"))) (err ERR-NOT-AUTHORIZED))
+    (asserts! (> (get ustx claim-entry) u0) (err ERR-NOTHING-TO-CLAIM))
+
+    (try! (as-contract (stx-transfer? (get ustx claim-entry) tx-sender sender)))
+    (try! (contract-call? .arkadiko-freddie-v1-1 deposit vault-id (get ustx claim-entry) reserve .arkadiko-token coll-type))
+
+    (if (is-eq stack-yield true)
```
philipdesmedt, line 80: @nieldeckx this will always fail unless the user does this when the stacker is in a cooldown cycle (i.e. after the 3 PoX cycles), but we pay out per PoX cycle.

Why? Check the `enable-vault-withdrawals` method in a stacker. We have the following assertion there:
```
    (asserts!
      (or
        (is-eq u0 (var-get stacking-stx-stacked))
        (>= burn-block-height (var-get stacking-unlock-burn-height))
      )
      (err ERR-BURN-HEIGHT-NOT-REACHED)
    )
```

Amount of STX stacked is not 0 and burn block height is not greater than the unlock burn height yet (since we stack for 3 PoX cycles). As a result, it will always throw `(err ERR-BURN-HEIGHT-NOT-REACHED)`.

We can keep this code, but it will only be useful for people claiming when we are in a cooldown.

philipdesmedt: Now that I look at the stacker, I think maybe we didn't even need that assertion, since a user's tokens are stacked anyway (they can't be withdrawn/unlocked), so it would be safe the let the user unlock the vault.

Too late for now either way.

nieldeckx: @philipdesmedt got it, thx for the info.

nieldeckx (approved):

PR #322, opened 2021-10-29
feat: upgrade governance to include a DAO fast track for proposal

philipdesmedt:
The Core Contributors  would like to propose a Governance Proposal to add a secondary upgrade mechanism to the DAO. The current Governance mechanism consists of making a proposal on-chain and then waiting for the result after a 10 day period. This is a good mechanism for long-term changes that need sufficient time for everybody to respond and cast their vote. 

However, that long wait period of 10 days proved to be the cause of delay when we wanted to bring back Swap v2 after the exploit. While the fix was written in code days after the exploit, the fix could only be implemented after the Governance vote had passed. This is a consequence of going the decentralized route where even the team does not have full control of the deployed code base and only on-chain governance can make changes.

We would like to propose a secondary mechanism through which the team can propose a change with a shorter running time. Only the DAO-contract will be able to create the vote (whereas with regular Governance, anyone with 1% of the DIKO supply can make a proposal). Currently, the Core Contributors have the keys the DAO (we can discuss on how to decentralize the access in the future). So in effect, only the active development team would be able to create proposals in this faster governance process. 

This governance fast-track should only be used to fix bugs or exploits. Make changes to existing functionality from a functional perspective, eg. fixing a mistake or unintended behavior. It should never be used to add new features, change important parameters or as a general alternative for the regular governance process.

The new time for the governance vote would be set at 250 blocks, roughly 40 hours. This strikes a good balance between giving enough time to vote and still being able to react swiftly in case it really needs to happen. 

We will create the on-chain governance vote for this as soon as possible. 

--- 

#### Addition 1

This governance fast-track should only be used to fix bugs or exploits. Make changes to existing functionality from a functional perspective, eg fixing a mistake or unintended behaviour. It should never be used to add new features, change important parameters or as a general alternative for the regular governance process.


nieldeckx (approved):
Looks good. Have added some extra tests.

kcdenman:
There are no limits on who can vote on these proposals, correct? The full community votes, it's just a shortened window of 40 hours instead of 10 days?

philipdesmedt:
@kcdenman correct

PR #325, opened 2021-11-02
tweak: update balances table with more info

Macxim:
Screen Shot 2021-11-02 at 7 02 55 AM

Feel free to adjust copy.

Also, it would be nice to have a `—` when data is not applicable. For example: 
Screen Shot 2021-11-02 at 7 07 01 AM

I managed to do it for STX Balance but not for the rest of them.


philipdesmedt (approved):
Great improvements!

I will merge this and we'll do another iteration later

PR #327, opened 2021-11-02
tweak: Add Alert component

Macxim:
Implementing proper Alert component to replace bits of HTML scattered here and there in the codebase.
Less code = less bugs.


philipdesmedt (approved):
Great improvement.

Issue #340, opened 2021-11-12
Swap v2: If the amount is more than 999 DIKO, it gets back to zero.


Macxim:
https://user-images.githubusercontent.com/1909957/141437759-3d79728a-6bc2-4673-9c25-f55fb119a7e0.mov

PR #341, opened 2021-11-12
fix: show token Y amount as text in swap

philipdesmedt:
This is a fix for #340 

Not sure if it's the right one, but it fixes the issue.


nieldeckx (approved):
It works.
(can remove the log)

Macxim:
It fixes it but I don't get why though 😕

Macxim:
I think it may be because `toLocaleString` returns a string, and the `type="number"` of the `input` doesn't like it.

Macxim (approved):

PR #350, opened 2021-11-15
tweak: add info when adding extra collateral

### LINKED ISSUE #344: Add education when depositing in a vaults that is stacking
funds are not stacked but are locked

Macxim:
Closes #344 

Feel free to update copy.

Do we need to display the alert conditionally?

Screen Shot 2021-11-15 at 8 01 27 PM


philipdesmedt (approved):
We can keep it static. Too important to do conditionally

PR #357, opened 2021-11-16
Swap improvements

nieldeckx:
1) Ability to enter amount for token Y
2) Disable button if insufficient balance

Note:
When entering an amount for token Y, we calculate the amount for token X and use it to perform the swap. There is no guarantee that you will receive the exact amount as entered for token Y because our swap contract does not support such feature. So not really sure if it's a good idea to allow this feature. Thoughts?


Macxim:
Like you mentioned, IMO we would need to update the swap contract before enabling this in the front-end. It might bring confusion among users to receive a different amount that what they expect.

philipdesmedt:
@nieldeckx well, if you are filling in token Y, what you are doing is you are taking the `swap-y-for-x` trade (instead of the `swap-x-for-y` trade or vice versa), no? So technically we have this functionality in the contract? Or at least something that approximates it closely?

nieldeckx:
@philipdesmedt That was not really the idea. Both `swap-x-for-y` and `swap-y-for-x` take dx as input, but it does not guarantee an exact output. On Uniswap you can enter the exact output amount for a swap and the input would be dynamic.

philipdesmedt:
@nieldeckx I see. Yes, then we can just leave out that feature. I'd rather not have more confusion on the swap (leads to more support)

nieldeckx:
Removed the code to enter token Y. So this branch includes:
1) Usage of swap v2 in te codebase
2) Button shows "insufficient balance" if applicable

Macxim:
There is another occurrence for `swap-v1` in `arkadiko/scripts/add-contract.js`. Does it need to be changed?

Macxim (approved):
Just had one comment, apart from that, good for me. Good job!

nieldeckx:
> There is another occurrence for `swap-v1` in `arkadiko/scripts/add-contract.js`. Does it need to be changed?

No need as it's not the swap itself but a related contract which is not part of the protocol.

PR #360, opened 2021-11-17
fix: swap UI to reflect calculations of contract

### LINKED ISSUE #343: Check if slippage set is high enough for transaction to succeed
We can check the price impact in the FE

Macxim:
Closes #343 

https://user-images.githubusercontent.com/1909957/142399122-3fb870ef-a201-425a-bf1e-c7109274a077.mov


Macxim:
I made some other changes that require to be carefully reviewed: https://github.com/arkadiko-dao/arkadiko/pull/360/commits/36aac4b7edcbe56e134d20e3a3b1d4fad427bed0#diff-39d2ae13b51c9e44d4f3acf60c8cd79ad6368abe6f0e0b69e0cf2c457550d201

Especially the replacement of `parseInt` with `Math.floor`. In general we shouldn't use `parseInt` with `number` since it expects a `string`

nieldeckx:
@philipdesmedt While playing around with the swap on mocknet, I noticed that the output amount we show in the UI is not correct. As there are no other swaps happening on mocknet, the output amount we show in the UI should be the exact same as what the user gets from the contract.

**Setup:**
Created wSTX/DIKO pair:
- wSTX: 223
- DIKO: 1000

**Calculations:**
Based on the logic in the swap smart contract.
Screenshot 2021-11-18 at 14 36 49

**Test swap**
UI is now showing the amount that the user actually gets after swapping.

Screenshot 2021-11-18 at 14 37 15

Screenshot 2021-11-18 at 14 22 00

**Price impact**
Before the swap: 1 DIKO = 0.223 STX
After the swap: 1 DIKO = 0.2434 STX
So price impact is 9.14%, which is now correctly shown in the UI.

**Slippage**
Now that we have the exact output amount, we can take into account the slippage set by the user. Slippage setting is used to calculate the minimum amount received. In the example:
- Theoretically we would receive: 42.79521 DIKO
- User has set 4% slippage
- So minimum amount to receive = 42.79521 DIKO * 96% = 41.083401
This is now shown in the UI. As we can not predict price changes that will happen, I don't think we can predict if the transaction would fail because of slippage.

philipdesmedt:
@nieldeckx ok - so what do you suggest we do? I understand your reasoning I think, but don't understand conclusion/next steps?

nieldeckx:
@philipdesmedt I updated the code already. Just wanted to show why it was updated + that the idea of checking slippage before swapping is not possible. PR is ready for review.

Macxim:
Maybe we need to change the name of the PR then?

philipdesmedt:
OK - I'll review.

philipdesmedt (approved):

Issue #299, opened 2021-10-23
Transaction history list does not show all pending transactions


nieldeckx:
Only 96 mempool transactions are taken into account. We should go over all result pages.

PR #361, opened 2021-11-18
fix: fetch all mempool pages to find pending transactions for a user

nieldeckx:
Fix for: https://github.com/arkadiko-dao/arkadiko/issues/299

### DIFF
--- a/web/common/transactions.ts
+++ b/web/common/transactions.ts
@@ -23,14 +23,28 @@ export const getAccountTransactions = async (address: string, contractAddress: s
 export const getPendingTransactions = async (address: string, contractAddress: string) => {
   const config = new Configuration({ basePath: network.coreApiUrl });
   const api = new TransactionsApi(config);
-  const txs = await api.getMempoolTransactionList({ limit: 96 });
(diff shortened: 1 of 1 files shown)


web/common/transactions.ts
```diff
@@ -23,14 +23,28 @@ export const getAccountTransactions = async (address: string, contractAddress: s
...
-      tx.tx_status === 'pending'
-  );
 
-  return list as MempoolContractCallTransaction[];
+  // Find amount of pages
+  const result = await api.getMempoolTransactionList({ limit: 1 });
+  let pages = Math.ceil(parseFloat(result.total) / 200.0);
+
+  // Go over all pages
+  var swapTransactions = [];
```
philipdesmedt, line 32: REQ: use `let` for variable definitions. Use `const` for variables that don't change. I think you can use `const` here.

philipdesmedt (changes requested):

web/common/transactions.ts
Macxim: `yarn lint:fix && yarn lint:prettier:fix` and you're good to go.

philipdesmedt (approved):

PR #362, opened 2021-11-18
Replace old modal with new component

Macxim:
I extracted the modals from the `manage-vault` component to their own components.
This requires a very thorough review as we don't want modals to be broken (again) and prevent users from managing their vaults.

Another thing that I changed is showing the Mint button only if availableCoinsToMint > 0.


philipdesmedt:
I'll review soon 🤞

philipdesmedt:
There are some small bugs with this that I need to fix, e.g. the amount to withdraw always shows 0 STX.

PR #369, opened 2021-11-20
Use parallel network requests

dmitriyshx:


philipdesmedt (approved):

philipdesmedt:
@dmitriyshx thanks, great work!

@nieldeckx do you want to merge this? I think we should do this before your other performance improvement on the stake page, since it will result in merge conflicts. I'll let you merge.

nieldeckx:
@dmitriyshx thanks a lot! 🙌

PR #372, opened 2021-11-22
feat: add governance v2 references in UI

philipdesmedt:
Deploying the contract.. will update description when it's deployed.

When we update governance to v2, we need to make sure everyone has withdrawn their DIKO from the first governance contract though... since it might be locked.


philipdesmedt:
Deployed here: https://explorer.stacks.co/txid/0x44bc20ebd14a8641cc461c20bac939403eb7710b0cad9f33bfe9a831b5894405?chain=mainnet

nieldeckx:
@philipdesmedt made sure the UI works with both governance V1 and V2 for now. We can exclude V1 once everyone has withdrawn their votes/tokens from V1.

nieldeckx (approved):
Looks good.

philipdesmedt:
@nieldeckx right - I will have to fix that.. should I just hardcode that if proposal ID is 1 or 2, it claims on the v1 contract?

philipdesmedt:
@nieldeckx okay, missed that... thanks!

PR #374, opened 2021-11-22
fix: DIKO pool rounding error

nieldeckx:
**Migration**
1) Add a governance proposal to replace the previous pool. 
2) Just before ending the proposal, call `add-rewards-to-pool` so all rewards are added to the old pool
3) End the proposal. This makes sure the old pool can't be used anymore as minting/burning will fail when staking/unstaking.
4) Migrate DIKO from old pool to new pool
5) Update the `last-reward-add-block` to align with the block from step 2. This makes sure no rewards are missed.
6) Add new pool to stake registry to make the new pool active

This makes sure:
- We properly shut down the old pool and activate the new one. Users won't be able to stake/unstake in the wrong pool.
- Users do not need to migrate
- All rewards are distributed

**Cooldown**
We ignore cooldown of pool V1. Users need to start cooldown on V2 again.


philipdesmedt:
@nieldeckx imho we can ignore the cooldown of pool v1 (so first option)

philipdesmedt:
@nieldeckx one thing before reviewing, could you make it easier to see what changed? you created a whole new file, but if I could see the diff, it would be easier to verify the bugfix

nieldeckx:
@philipdesmedt done!

philipdesmedt (approved):
LGTM.

We can discuss and come up with a deployment plan once the current DAO upgrade governance vote has ended.

philipdesmedt:
I updated the front-end in this PR.

We should not merge this until the contract is deployed and the new pool is active.

PR #384, opened 2021-11-26
fix: Add DIKO column and update modal with DIKO amount

### LINKED ISSUE #383: Unstake DIKO popup should show stDIKO balance
Unstake popup should focus on stDIKO.

Screen_Shot_2021-11-25_at_11 03 27_PM

Macxim:
Closes #383 

### Staking DIKO

Screen Shot 2021-11-26 at 4 03 24 PM

### Unstake modal
Screen Shot 2021-11-26 at 4 04 15 PM

### DIFF
--- a/web/components/unstake-diko-modal.tsx
+++ b/web/components/unstake-diko-modal.tsx
@@ -21,7 +21,6 @@ export const UnstakeDikoModal = ({
   showUnstakeModal,
   setShowUnstakeModal,
   stakedAmount,
-  stDikoToDiko,
 }) => {
   const [state, setState] = useContext(AppContext);
   const [errors, setErrors] = useState<string[]>([]);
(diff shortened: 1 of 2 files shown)


Macxim:
Not sure if DIKO amount is correct or if there is a better way to calculate it. @nieldeckx @philipdesmedt

web/components/unstake-diko-modal.tsx
```diff
@@ -107,7 +112,12 @@ export const UnstakeDikoModal = ({ showUnstakeModal, setShowUnstakeModal, staked
           minimumFractionDigits: 2,
           maximumFractionDigits: 6,
         })}{' '}
-        stDIKO.
+        stDIKO which equals to{' '}
+        {microToReadable(stakedAmount * stDikoToDiko).toLocaleString(undefined, {
```
philipdesmedt, line 116: I think this is fine, not sure if there can be slight rounding errors or an issue when a user unstakes & withdraws and gets less than the UI showed.... probably not?

philipdesmedt (approved):

nieldeckx (changes requested):
I don't think this is correct:
Screenshot 2021-11-29 at 08 58 35
I have 10 stDIKO which should be equal to ~11.9 DIKO.

The function `get-stake-of` on the pool will return the amount of DIKO the user should get, not the amount of stDIKO. For the amount of stDIKO you can just use the state.balances.

Macxim:
@nieldeckx Agree. I had the same thought yesterday. I'll change it in a bit.

Macxim:
Fixed. What do you think @nieldeckx?

Updated screenshots below:

#### Staking DIKO

Screen Shot 2021-11-29 at 8 41 07 AM

#### Unstake modal

Screen Shot 2021-11-29 at 8 40 51 AM

nieldeckx (approved):

nieldeckx:
@Macxim Perfect :)

PR #389, opened 2021-11-30
feat: analytics: add cryptocurrencies prices

Macxim:


analytics-web/components/pool-row.tsx
```diff
@@ -2,9 +2,11 @@ import React, { useEffect, useState } from 'react';
 import { stacksNetwork as network } from '@common/utils';
 import { callReadOnlyFunction, stringAsciiCV, cvToJSON } from '@stacks/transactions';
 import axios from 'axios';
+import { tokenList } from '../../web/components/token-swap-list';
```
philipdesmedt, line 5: Just a note.. I'm not sure if this will work when I create a production build since they are two separate projects. Will try building this branch and let you know.

Macxim: Yeah, I was wondering the same thing but I don't see why it wouldn't work. Worth trying.

philipdesmedt:
It seems to work but the placeholders for loading aren't showing.. not sure why. Gonna merge and we can look at this in another PR

PR #393, opened 2021-12-02
feat: when burning USDA, subtract stability fee from user input

### LINKED ISSUE #376: When burning USDA it's not clear the user needs to have enough USDA in wallet to also pay stability fee
User has 630.4 USDA in wallet. Tries to pay back 630 USDA but it fails:
https://explorer.stacks.co/txid/0xc2d6612c4ebd285bd214b61a45aecc9f4380871e7c5aae1f39666a2dfc08f259?chain=mainnet

The burn function will first pay back the stability fee before burning USDA. 

Solution might be to take the stability fee into account in the UI. Instead of using 630 USDA as debt parameter, we can use 630 minus the stability fee needing to be paid.

Macxim:
Closes #376


web/components/manage-vault.tsx
```diff
@@ -291,7 +291,7 @@ export const ManageVault = ({ match }) => {
       functionName: 'burn',
       functionArgs: [
         uintCV(match.params.id),
-        uintCV(parseFloat(usdToBurn) * 1000000),
+        uintCV(parseFloat(usdToBurn - stabilityFee) * 1000000),
```
nieldeckx, line 294: The stabilityFee comes from the contract so the integer includes 6 digits. While usdToBurn comes from user input so does not include 6 digits. I think it needs to be: uintCV((parseFloat(usdToBurn) * 1000000) - stabilityFee)

As we are now subtracting the stability fee from user input we can also optimise the post condition a bit:
let totalToBurn = Number(usdToBurn) + 2 * (stabilityFee / 1000000);
Can become:
let totalToBurn = Number(usdToBurn) + (stabilityFee / 1000000);

web/components/manage-vault.tsx
```diff
@@ -968,7 +968,7 @@ export const ManageVault = ({ match }) => {
                 <div className="mt-2">
                   <p className="text-sm text-gray-500">
                     Choose how much USDA you want to burn. Burning will include a stability fee of{' '}
-                    {stabilityFee / 1000000} USDA, so take this into account.
+                    <span className="font-semibold">{stabilityFee / 1000000} USDA</span>, so take this into account.
```
nieldeckx, line 971: We can remove "so take this into account" as our frontend logic is now taking it into account.

nieldeckx (changes requested):

web/components/manage-vault.tsx
Macxim: Yeah. Good catch. I'll update it.

Macxim:
All yours @nieldeckx

nieldeckx (approved):
Looks good. Tested on prod and got the following result, which looks good:

**Prev:**
Stability fee: 0.759696
User enters: 198.931394
Contract call: 198.931394

**New:**
Stability fee: 0.759696
User enters: 198.931394
Contract call: 198.171698 (= 198.931394 - 0.759696)

philipdesmedt:
@nieldeckx what if it takes longer than 1 block to mine this TX? That is not an outlier but happens all the time.. the stability fee goes up per block, so this is no longer correct then and the TX will fail?

nieldeckx:
@philipdesmedt Correct. We could calculate the stability fee for next 10 (?) blocks and subtract this as well?
If the transaction gets mined in the next block, the user will have some USDA left.

philipdesmedt:
@nieldeckx yeah, that's probably best to do. 10 blocks is fine

philipdesmedt:
@Macxim are you going to fix the above? or should someone else?

Macxim:
@philipdesmedt Since it's quite high priority, I guess either of you guys can go ahead with it.

philipdesmedt:
@nieldeckx @Macxim I just subtracted 1.5 times the stability fee now

PR #396, opened 2021-12-07
fix: upgrade auction engine

philipdesmedt:


philipdesmedt:
@nieldeckx ready for review

philipdesmedt:
To deploy this I will also have to deploy a new trait.. looking into that now.

PR #401, opened 2021-12-14
tweak: remove vault rewards

Macxim:
Screen Shot 2021-12-14 at 11 26 46 AM

Screen Shot 2021-12-14 at 11 26 52 AM


philipdesmedt (changes requested):
We should still allow people to claim rewards for at least a while, but we can change the message.

Macxim:
Ah yeah you're right! 😅

Macxim:
Screen Shot 2021-12-14 at 11 42 12 AM

PR #408, opened 2021-12-28
feat: add STX/WELSH trading pair

philipdesmedt:
This PR adds the WELSH token to Arkadiko Swap (STX/WELSH).

The WELSH listing is subject to a decentralised governance vote.

The Welshcorgiteam will provide 45,000 STX liquidity to start trading.


web/components/token-swap-list.tsx
```diff
@@ -41,6 +41,15 @@ export const tokenList = [
     address: 'SP3DX3H4FEYZJZ586MFBS25ZW3HZDMEW92260R2PR',
     decimals: 8
   },
+  {
+    id: 4,
```
Artiume, line 45: increment?

Macxim: I'm guessing @philipdesmedt isn't done with it yet.
Also the logo path is wrong on L48. Thanks for  having a look though 👀

PR #416, opened 2022-01-07
feat: improve speed of Stake page

philipdesmedt:


Macxim:
Awesome! Really good job. 🚀 
Can you maybe run `prettier "./components/stake.tsx" --write` when you're done?

Macxim (approved):

philipdesmedt:
@Macxim done, I added a local binary for prettier so we can run `./node_modules/prettier/bin-prettier.js "./components/stake.tsx" --write` in the web folder

Macxim:
`prettier` was already in devDependencies though, so you could just run `yarn lint:prettier:fix` and it should have linted all files.
We could also add a pre-commit hook for this.

philipdesmedt:
@Macxim right, I forgot :D I will remove this again

PR #419, opened 2022-01-20
feat: use API to optimise Swap

philipdesmedt:


philipdesmedt:
@Macxim @nieldeckx can you review this?

nieldeckx (approved):
Great! From 6s to 1s :)

Macxim (approved):
I couldn't find any issue and this feels much smoother 🚀 :shipit:

PR #420, opened 2022-01-20
feat: add xBTC USDA minting

philipdesmedt:


clarity/contracts/arkadiko-auction-engine-v1-1.clar
```diff
@@ -258,12 +258,16 @@
 ;; @desc calculate the discounted auction price on the (dollarcent) price of the collateral
 ;; @param price; the current on-chain price
 ;; @param auction-id; the ID of the auction in which the collateral will be sold
-(define-read-only (discounted-auction-price (price uint) (auction-id uint))
+(define-read-only (discounted-auction-price (price uint) (decimals uint) (auction-id uint))
   (let (
     (auction (get-auction-by-id auction-id))
     (discount (* price (get discount auction)))
+    (total-decimals (if (> decimals u0)
```
nieldeckx, line 265: Is this just to make sure or did you find an actual use case for this?

nieldeckx (commented):
Method 'calculate-usda-count' in sip10-reserve does not take into account decimals, why?

clarity/contracts/arkadiko-auction-engine-v1-1.clar
philipdesmedt: If the oracle cannot find a price for the token passed, it returns 0 for price and 0 for decimals. Price can be 0 (then it just returns 0 I think). But for decimals, you would divide by 0 and it crashes

philipdesmedt:
@nieldeckx about your comment (`Method 'calculate-usda-count' in sip10-reserve does not take into account decimals, why?`)... I think it's actually a bug in the stx and sip10 reserves, but it is being blocked by Freddie on line 322 and 554 in arkadiko-freddie-v1-1. Basically I think we calculate the ratio in freddie, and then the max USDA mintable. The max USDA calaculation is wrong, I think? Can you verify this?

philipdesmedt:
@nieldeckx I don't think it was needed, but I also fixed the bug in the sip10 reserve since we'll be deploying a new one

nieldeckx:
Great work, looks good to me.

PR #421, opened 2022-01-21
feat: lower governance proposal requirements and require minimum participation

philipdesmedt:


clarity/contracts/arkadiko-governance-v2-1.clar
```diff
@@ -245,14 +248,14 @@
         (contract-of stake-pool-diko)
         (unwrap-panic (contract-call? .arkadiko-dao get-qualified-name-by-name "stake-pool-diko"))
       )
-      (err ERR-NOT-AUTHORIZED)
+      (err ERR-WRONG-POOL)
```
nieldeckx, line 251: Suggest to use the same error in 'propose' and 'vote-against' as well.

nieldeckx (commented):
Looks good!

scripts/governance-add-proposal.js
```diff
@@ -16,12 +16,12 @@ async function transact() {
       tx.uintCV(43100),
       tx.uintCV(900),
       tx.stringUtf8CV('Add STX/WELSH Pair'),
```
Scotty0448, line 18: I notice that this still says 'Add STX/WELSH Pair'.

philipdesmedt: Fixe

PR #433, opened 2022-02-02
tweak: separate governance v1 from current proposals

Macxim:
Not very beautiful code-wise 😞 but UI looks better and less confusing.
Screen Shot 2022-02-02 at 11 59 36 AM


philipdesmedt (approved):
Nice!

PR #450, opened 2022-02-23
feat: add multi-hop swap contract

philipdesmedt:
Possible pairs simplifications:
1. STX/xBTC
STX -> USDA (swap x for y)
USDA -> xBTC (swap y for x)

2. STX/DIKO
STX -> USDA (swap x for y)
USDA -> DIKO (swap y for x)


philipdesmedt:
@nieldeckx can you review the smart contract?

will adapt the UI after the SC is deployed

clarity/contracts/arkadiko-multi-hop-swap-v1-1.clar
```diff
@@ -0,0 +1,37 @@
...
+  (min-dz uint)
+  (inverse-first bool)
+  (inverse-second bool)
+)
+  (let (
+    (swapped-amounts (if inverse-first
+      (unwrap! (contract-call? .arkadiko-swap-v2-1 swap-y-for-x token-x-trait token-y-trait dx u0) (err ERR-SWAP-ONE-FAILED))
+      (unwrap! (contract-call? .arkadiko-swap-v2-1 swap-x-for-y token-x-trait token-y-trait dx u0) (err ERR-SWAP-ONE-FAILED))
+    ))
+    (y-amount (unwrap-panic (element-at swapped-amounts u1)))
```
nieldeckx, line 27: Q: The function `swap-y-for-x` returns `(list dx dy)` where `dy` is the input and `dx` is the result. So we should get the first element from the list instead of the second?

nieldeckx (changes requested):
Just one question/remark.

clarity/contracts/arkadiko-multi-hop-swap-v1-1.clar
philipdesmedt: Yes that's a good point, I also need to do it conditional here. Will also write a test

philipdesmedt:
This has been deployed a while ago. Gonna merge this. Front-end is pending in another PR.

PR #453, opened 2022-03-02
tweak: rework new vault creation wizard

Macxim:


philipdesmedt:
@Macxim is this ready for review? can you fix merge conflicts

Macxim:
Fixed the conflicts. You can review.
A couple of things we need to fix:

- Add a `Back` button on Step 2
- The URL `http://localhost:9000/vaults/new?type=STX-A&token=STX` doesn't really lead to the right step anymore (like it did before, but instead goes to the choice between STX and xBTC)
- The checkbox should be fixed on Step 2
- Front-end checks would be cool to have on Step 2 upon clicking on Continue (instead of disabling the button)
- Clicking on step 1 of the nav `/vaults/new` doesn't seem to go back

philipdesmedt:
@Macxim I simplified this a little bit, think it's ready for merge

PR #456, opened 2022-03-08
Update README.md

moodmosaic:
Format error codes so that the show up in a code block, otherwise they appear all in the same line.


nieldeckx:
@moodmosaic thx!

PR #470, opened 2022-03-24
feat: add multihop UI

philipdesmedt:


philipdesmedt:
This is nearly ready.

@Macxim could you add some design on the path here? The only multihop pair will be STX/DIKO at launch, see screenshot below. I'm thinking of "logo <TOKEN_TICKER> --> logo2 <TOKEN_TICKER2> --> logo3 <TOKEN_TICKER3>"

image

nieldeckx (approved):
Tested locally and seems to work. Great work! PR is in draft, is there anything left to do?

philipdesmedt:
@nieldeckx don't think so, should be ready. Will update the status

PR #473, opened 2022-03-25
feat: add contract for ALEX/DIKO dual yield

philipdesmedt:


clarity/contracts/arkadiko-alex-dual-yield-v1-1.clar
```diff
@@ -0,0 +1,74 @@
...
+  )
+)
+
+
+
+;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;
+;; public methods (callable by ALEX contract or Arkadiko DAO) ;;
+;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;
+
+(define-public (mint (amount-of-blocks uint))
```
nieldeckx, line 36: Not sure, but I think Alex needs a mint method like:
`(define-public (mint (amount uint) (recipient principal))`

In the mint method we should:
- Mint DIKO and assign to contract (diko-to-mint = blocks-passed * get-rewards-per-block-for-alex)
- Send DIKO based on amount and recipient parameters

This way we are in full control of minted DIKO.

nieldeckx (commented):
Did a quick first review. Once we are sure the interface is correct, some tests would be nice :)

clarity/contracts/arkadiko-alex-dual-yield-v1-1.clar
philipdesmedt: @fiftyeightandeight I will let you comment on this. Let us know what you would need exactly or if this works

fiftyeightandeight: `(amount uint) (recipient principal)` would definitely help. Thank you.

philipdesmedt: @fiftyeightandeight ok I will look at it and refactor!

fiftyeightandeight: Hi @nieldeckx we need to pre-agree to the exact exact emission schedule, which will then be hard-coded into the relevant contracts.

philipdesmedt: @fiftyeightandeight I think what Niel means is that 1.25M DIKO vs 5.5% of our emissions are not exactly the same (could differ with a few thousand DIKO)... so to get exactly 1.25M DIKO would be very hard for us to calculate as we approximate it with a percentage and then there is a "rounding error" of a few basis points

I think what you could do on your end is call `get-rewards-per-block-for-alex` * `number-of-blocks` to get the exact amount of DIKO for the staking period you're giving out rewards for?

fiftyeightandeight: Ok, that should work too. So `mint` will accept, perhaps, `(amount-of-blocks uint)` and `(recipient principal)` as input?

fiftyeightandeight: Actually, `mint` will mint all the rewards for the number of blocks, correct? I.e. If the number of blocks was 525 blocks, then calling `mint` will mint 525 x 5.5% of your emission?

fiftyeightandeight: > Not sure, but I think Alex needs a mint method like: `(define-public (mint (amount uint) (recipient principal))`
> 
> In the mint method we should:
> 
> * Mint DIKO and assign to contract (diko-to-mint = blocks-passed * get-rewards-per-block-for-alex)
> * Send DIKO based on amount and recipient parameters
> 
> This way we are in full control of minted DIKO.

If you provide implementation like the above, think we can handle the amount to send to each `recipient` based on total DIKO minted to the contract up to that point.

philipdesmedt: @fiftyeightandeight yeah, ok.. we are going to simplify things a bit on our side.

We'll make a method `(define-public (mint (amount uint) (recipient principal))`. We have 1.25M DIKO per year and approximately 100 ALEX cycles per year (144 * 365 / 525 = 100.114 cycles). We will just mint 1.25M DIKO on the contract (not all at once but through a mint function we can call) and you can assume you can emit 12,500 DIKO per cycle (1.25M / 100). This way you can just pro-rata call the `mint` function per recipient.

Would that be fine? Sorry for the back and forth but I think this will be the easiest where both sides have to do the least calculations.

clarity/contracts/arkadiko-alex-dual-yield-v1-1.clar
```diff
@@ -0,0 +1,75 @@
...
+    (is-eq (unwrap-panic (contract-call? .arkadiko-dao get-emergency-shutdown-activated)) false)
+    (is-eq (var-get shutdown-activated) false)
+  )
+)
+
+;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;
+;; public methods (callable by ALEX contract or Arkadiko DAO) ;;
+;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;
+
+(define-public (mint (amount uint) (recipient principal))
```
fiftyeightandeight, line 25: @philipdesmedt given what we just discussed, I see that all we need to do is to call at each cycle  `mint (amount uint) (recipient principal)` with `amount` <= 12.5k (with the expectation being the sum of `amount` across all recipients for a given cycle would equal to 12.5k - this expectation will be codified into our contract).

clarity/contracts/arkadiko-alex-dual-yield-v1-1.clar
```diff
@@ -0,0 +1,75 @@
...
+;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;
+;; public admin methods       ;;
+;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;
+
+(define-public (mint-diko (number-of-cycles uint))
+  (begin
+    (asserts! (is-eq tx-sender (contract-call? .arkadiko-dao get-guardian-address)) (err ERR-NOT-AUTHORIZED))
+    (asserts! (is-activated) (err ERR-NOT-ACTIVATED))
+    (asserts! (<= number-of-cycles u10) (err ERR-TOO-MUCH-DIKO))
+    (asserts! (< (var-get diko-minted) MAX-TO-MINT) (err ERR-TOO-MUCH-DIKO))
```
fiftyeightandeight: `<=` makes more sense if we want to use `number-of-cycles` == 100?

clarity/contracts/arkadiko-alex-dual-yield-v1-1.clar
philipdesmedt: @fiftyeightandeight correct - each cycle you can give away 12.5K DIKO (12.5K DIKO if there would be only 1 LP).

Each LP should get pro rata an amount of DIKO vs their LP tokens staked and the sum of these amounts per cycle is 12.5K DIKO

clarity/contracts/arkadiko-alex-dual-yield-v1-1.clar
philipdesmedt: Yeah - you're right. I updated it.

clarity/contracts/arkadiko-alex-dual-yield-v1-1.clar
fiftyeightandeight: @philipdesmedt for 12.5k DIKO, `amount` should be 12.5k * 1e6 (i.e. 6 decimals for DIKO), correct?

philipdesmedt: @fiftyeightandeight yes, correct! Always pass it with 6 decimals

philipdesmedt:
@nieldeckx @fiftyeightandeight I think this should be ready for deployment

fiftyeightandeight:
> @nieldeckx @fiftyeightandeight I think this should be ready for deployment

Looks good to me!

nieldeckx:
> @nieldeckx @fiftyeightandeight I think this should be ready for deployment

Looks good!

PR #475, opened 2022-04-06
tweak: update new emissions

philipdesmedt:
This PR does a few things:

- Calculate APR according to new emissions
- Add scripts to update emissions on stake registry

### DIFF
--- a/scripts/set-pool-data-xbtc.js
+++ b/scripts/set-pool-data-xbtc.js
@@ -14,10 +14,8 @@ const txOptions = {
     tx.stringAsciiCV('wSTX-xBTC LP'),
     tx.uintCV(0),
     tx.uintCV(0),
-    tx.uintCV(50000)
+    tx.uintCV(0)
   ],
-  nonce: new BN(437, 10),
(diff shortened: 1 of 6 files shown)


philipdesmedt:
@nieldeckx can you do a quick review? Mostly check if the set-pool-data calls in the scripts are correct

scripts/set-pool-data-xbtc.js
```diff
@@ -12,12 +12,10 @@ const txOptions = {
   functionArgs: [
     tx.contractPrincipalCV(CONTRACT_ADDRESS, 'arkadiko-stake-pool-wstx-xbtc-v1-1'),
     tx.stringAsciiCV('wSTX-xBTC LP'),
+    tx.uintCV(55200), // completely disable
```
nieldeckx, line 15: To be fully correct you would also need to set "deactivated-rewards-per-block" to the amount of rewards per block on block "deactivated-block". But it's not really needed as most rewards are already added.

nieldeckx (commented):
Looks good. One minor comment.

PR #477, opened 2022-04-20
fix: liquidation pool post conditions

nieldeckx:


web/components/liquidations-reward.tsx
```diff
@@ -29,6 +34,42 @@ export const LiquidationReward: React.FC<LiquidationRewardProps> = ({
   const [state, setState] = useContext(AppContext);
 
   const claim = async () => {
+
+    const postConditions = [];
+    if (tokenIsStx) {
+
```
philipdesmedt, line 40: NTH: remove newline

web/components/liquidations-reward.tsx
```diff
@@ -29,6 +34,42 @@ export const LiquidationReward: React.FC<LiquidationRewardProps> = ({
...
+          FungibleConditionCode.Equal,
+          uintCV(claimable).value,
+        )
+      )
+    } else {
+
+      // FT name
+      var tokenName = "diko";
+      Object.keys(tokenTraits).forEach((key, index) => {
+        if (tokenTraits[key].address == token.split('.')[0] && tokenTraits[key].swap == token.split('.')[1]) {
```
philipdesmedt, line 55: NTH: use `filter` instead of a forEach and if statement

```
Object.keys(tokenTraits).filter(obj => obj.address === token...
```
something like that

philipdesmedt (changes requested):

PR #480, opened 2022-04-21
Upgrade dependencies

Macxim:
I upgraded pretty much all the dependencies I was able to.
I didn't touch anything related to Webpack though.

The only library that is giving me issues is `@stacks/transactions`.
This can probably be done in another PR.

Could you please review and check everything is still working accordingly?

Thanks!


nieldeckx (changes requested):
App is not working here:
Screenshot 2022-04-25 at 14 49 38

Looks like something changed when returning a tuple. Haven't found the solution yet.

philipdesmedt:
@nieldeckx @Macxim just tried this and works for me now... I did have to do use the standard Stacks Node API URL tho (and not the Syvita or another one).. you can remove it using `localStorage.removeItem('arkadiko-stacks-node')` to fall back on the standard URL if you're using a custom one

philipdesmedt:
could you guys review?

nieldeckx (commented):
It works for me. Tested a few transactions on mainnet.

Did notice that not all packages are using the latest version, for example:
`"@stacks/transactions": "3.2.0",`
Not sure if important..

philipdesmedt:
@nieldeckx yeah, good point.

Let's merge this and upgrade @stacks/transactions to 4.3.0 in another PR (will do that tomorrow), https://www.npmjs.com/package/@stacks/transactions

PR #485, opened 2022-05-09
Add pill to header Governance item when a vote is in progress

Macxim:
Screen Shot 2022-05-09 at 12 52 24 PM

### DIFF
--- a/web/components/header.tsx
+++ b/web/components/header.tsx
@@ -1,4 +1,4 @@
-import React, { useContext } from 'react';
+import React, { useContext, useEffect, useState } from 'react';
 import { Disclosure } from '@headlessui/react';
 import { AppContext } from '@common/context';
 import { NavLink as RouterLink } from 'react-router-dom';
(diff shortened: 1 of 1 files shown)


web/components/header.tsx
```diff
@@ -28,6 +32,49 @@ export const Header: React.FC<HeaderProps> = ({ signOut, setShowSidebar }) => {
...
   const name = bnsName();
+  const contractAddress = process.env.REACT_APP_CONTRACT_ADDRESS || '';
+  const stxAddress = useSTXAddress();
+  const [isVotingOpen, setisVotingOpen] = useState(false);
+
+  useEffect(() => {
+    let mounted = true;
+
+    const getData = async () => {
+      const proposals = await callReadOnlyFunction({
```
philipdesmedt, line 43: REQ: we don't need to fetch v2 proposals since there will never be an open one there

philipdesmedt (changes requested):

web/components/header.tsx
Macxim: Yeah. I wasn't sure. I'll change this. 👍

web/components/header.tsx
```diff
@@ -28,6 +32,37 @@ export const Header: React.FC<HeaderProps> = ({ signOut, setShowSidebar }) => {
...
+        contractAddress,
+        contractName: 'arkadiko-governance-v3-1',
+        functionName: 'get-proposals',
+        functionArgs: [],
+        senderAddress: stxAddress || '',
```
philipdesmedt, line 48: REQ: this page will fail when the user is not logged in (e.g. try incognito)

The fix is to remove this `stxAddress` completely (also line 36 + the import) and just use `contractAddress` as the sender address

philipdesmedt (changes requested):

web/components/header.tsx
Macxim: Right, good catch. Will update.

PR #486, opened 2022-05-09
fix: improve auction page API calls

nieldeckx:


philipdesmedt (approved):
This works, fantastic

PR #488, opened 2022-05-13
Small UX/UI improvements for Liquidations

Macxim:
I disabled the button if there are no redeemable STX to prevent useless transactions.

Screen Shot 2022-05-13 at 11 00 35 PM


nieldeckx (approved):
Looks better indeed :)

PR #490, opened 2022-05-15
tweak: rename sections

Macxim:
- auctions => liquidations (also change route path)
- vaults => borrow (didn't change route path)


web/components/routes.tsx
```diff
@@ -44,7 +43,7 @@ export function Routes() {
       <Route exact path="/swap/remove/:currencyIdA/:currencyIdB" component={RemoveSwapLiquidity} />
       <Route exact path="/pool" component={Pool} />
 
-      <Route exact path="/auctions" component={Liquidations} />
+      <Route exact path="/liquidations" component={Liquidations} />
```
philipdesmedt, line 46: REQ: put an additional route /auctions that also links to Liquidations.

philipdesmedt (changes requested):

PR #498, opened 2022-06-08
feat: add stacker payer contract to redeem xSTX

philipdesmedt:


philipdesmedt:
@nieldeckx can you review?

api/app/models/pox_calculator.rb
```diff
@@ -47,6 +47,44 @@ def self.calculate_usda_yields(names: ['stacker'], file: 'vaults-pox-33.json', u
...
+            total_amount += collateral_amount.to_i
+            stacked_amount += stck_amount.to_i
+          rescue
+            puts "Failed TX with id #{txid}"
+          end
+        end
+        sleep(2)
+      end
+
+      puts "Total Collateral: #{total_amount}"
```
nieldeckx, line 79: Q: this also takes non stacking vaults into account? For these we never minted xSTX so they don't need to be included?

clarity/contracts/arkadiko-stacker-payer-v2-1.clar
```diff
@@ -0,0 +1,74 @@
...
+
+(define-public (toggle-stacker-payer-shutdown)
+  (begin
+    (asserts! (is-eq tx-sender (contract-call? .arkadiko-dao get-guardian-address)) (err ERR-NOT-AUTHORIZED))
+
+    (ok (var-set stacker-payer-shutdown-activated (not (var-get stacker-payer-shutdown-activated))))
+  )
+)
+
+(define-public (return-stx-to-reserve (ustx-amount uint))
```
nieldeckx, line 33: Q: is this really needed or just in case something goes wrong?

web/components/liquidations.tsx
```diff
@@ -75,23 +76,43 @@ export const Liquidations: React.FC = () => {
...
-      onFinish: data => {
-        setState(prevState => ({
-          ...prevState,
-          currentTxId: data.txId,
-          currentTxStatus: 'pending',
-        }));
-      },
-      anchorMode: AnchorMode.Any,
-    });
+    if (redeemableStx > 0) {
```
nieldeckx, line 79: Suggestion: contract function to encapsulate both redeem-stx methods so user only needs one transaction to redeem.

clarity/contracts/arkadiko-stacker-payer-v2-1.clar
```diff
@@ -0,0 +1,74 @@
...
+    (if (> ustx-amount u0)
+      (as-contract
+        (stx-transfer? ustx-amount tx-sender (unwrap-panic (contract-call? .arkadiko-dao get-qualified-name-by-name "stx-reserve")))
+      )
+      (ok true)
+    )
+  )
+)
+
+(define-public (set-stx-redeemable (ustx-amount uint))
```
nieldeckx, line 46: Suggestion: wondering if we can decentralise this centralised part. The contract could have a public method which takes the vault-id as parameter and calculates the the difference between total collateral and stacked tokens. The calculated difference is added to stx-redeemable and the method can only be called once per vault.

nieldeckx (commented):
Did a first review. 
One thing I do not understand yet is that freddie mints xSTX equal to total STX collateral in vault. But not all collateral has to be sold in a liquidation. So there must be xSTX left which can not be claimed by users?

api/app/models/pox_calculator.rb
philipdesmedt: This also takes non-stacking vaults into account, yes. However, this script was just written to have a rough calculation. I'll improve it either way to exclude non-stacking liquidated vaults.

philipdesmedt:
@nieldeckx "So there must be xSTX left which can not be claimed by users?"... yes - I think this is correct and is likely the xSTX still in SIP10 reserve v2

clarity/contracts/arkadiko-stacker-payer-v2-1.clar
philipdesmedt: @nieldeckx that's fair, but requires a lot of on-chain transactions now to recover? We could initialize it to a correct value perhaps. I'll look into it

clarity/contracts/arkadiko-stacker-payer-v2-1.clar
philipdesmedt: Just in case something goes wrong that no STX are lost.

philipdesmedt:
@nieldeckx to elaborate on this, the following TX is a good example https://explorer.stacks.co/txid/0x9227bcecb300605843b840572fe26c988d4e6929fa49b9ed4c96414319afab02?chain=mainnet

1. 92 xSTX was minted (equivalent to amount of STX in vault)
2. Equivalent of 73 xSTX in USDA was burned (35 USDA), so we minted 92-73 xSTX (19 xSTX) too much. This is just sitting in SIP10 Reserve and cannot be redeemed.
3. Only 62 STX was allowed for redemption (amount of stacked-tokens)
4. We are missing 73-62 STX to redeem. This should be redeemed through the new stacker payer

nieldeckx:
So to be fully correct we need to go over all auctions to find out the collateral sold. With this info and the vault info we can calculate the amount of xSTX which was minted too much and the amount of xSTX which is immediately redeemable?

To solve this for the future we could update the auction-engine:
- auction-engine method `finalize-liquidation` to calculate the missing redeemable STX and xSTX which was minted too much
- auction-engine can burn xSTX which was minted too much
- auction-engine can use stacker-payer as a proxy to be able to redeem STX

philipdesmedt:
@nieldeckx yes - I have updated the Ruby script now and running the calculation

PR #499, opened 2022-06-09
fix: liquidation reward groups

nieldeckx:


philipdesmedt (changes requested):

nieldeckx:
@philipdesmedt should be good now.

An example of failed TX:
https://explorer.stacks.co/txid/0x2b031db1421bacc9f14c44dcdb7a8d027b336eb6087ad400b70b7d55615feaf3?chain=mainnet

Is fixed for this user as each reward ID is now only used once:
Screenshot 2022-06-09 at 13 51 20

PR #503, opened 2022-06-22
feat: add updated Oracle implementation

philipdesmedt:
This uses a double strategy:

1. if the token passed is a token known by redstone (i.e. STX or BTC), then use the RedStone Oracle to verify the message (i.e. price) passed
2. if the token passed is not a token known by redstone (i.e. DIKO, atALEX, USDA, ...), then use the custodial method of updating via private key


nieldeckx:
Closing this PR. Working on multisig oracle: https://github.com/arkadiko-dao/arkadiko/pull/524

PR #509, opened 2022-07-11
Add atALEX as collateral type

YannisDC:
Governance proposal to add atALEX as collateral type:

- Stability Fee: 4%
- Debt Ceiling: 250,000 USDA
- Liquidation Penalty: 20%
- LTV: 25%


philipdesmedt (changes requested):

nieldeckx:
FYI: in order for the liquidations to work we need to add the new token in `arkadiko-liquidation-rewards-v1-1` via the method  `update-reward-tokens`

philipdesmedt:
I think everything is ready here

@Macxim could you add some CSS for the right atALEX colors?

image

Macxim:
@philipdesmedt It was already there. Just needed to add the correct class. I did it 👍 Let's go 🚀

philipdesmedt:
@Macxim thanks!

@nieldeckx gonna merge this - are we still using another branch to make production builds?

nieldeckx:
@philipdesmedt just merged the new auction-engine branch. So pull in master and you should be good.

PR #512, opened 2022-08-03
feat: add borrow section with collateral types cards

Macxim:
What I did:

- design new cards for the new, more attractive Borrow section
- get user's balance for STX and xBTC to provide a better UX when indicating how much a user can borrow given a certain amount of collateral
- if user's balance is 0 or their wallet isn't connected, we want to show a specific number, otherwise it doesn't make sense (just a thought now that I think of it: maybe we should put another threshold instead of 0: let's say a user has 10 STX in their wallet, it may not be _that_ interesting to display how much a user can borrow with such a small amount) 
- add hash to Step one of Vault creation to make sure STX button is active and both STX-A and STX-B are visible when coming from the Borrow section after clicking on the STX card
- center cards for now, until we have another collateral type
- `Your vaults` has been renamed to `Your positions` (more on this later in another PR 👀)


Macxim:
@philipdesmedt @nieldeckx 

Could you help with the rest?
- calculations for when users have their wallet connected (L99)
- calculations for when users have *not* their wallet connected (I put `2000 STX` and `2 xBTC`, feel free to change)
- the page should be visible even  when users have *not* their wallet connected

nieldeckx:
@Macxim fixed the example calculations and the available liquidity on a card. As both depend on the collateral type, I added 2 cards again for STX (STX-A and STX-B collateral types).

When you showed the screenshots, the cards had a coloured background so they really popped out. But when I run the website they have a grey background:
Screenshot 2022-08-06 at 08 21 13

I guess we can get rid of this intermediate screen if we just add all info to the first screen? Only need to add the 'liquidation penalty' to the first screen. And maybe also the info icons with popups?
Screenshot 2022-08-06 at 08 24 12

Macxim:
@nieldeckx 
Good job!

Unless I'm mistaken I don't think we want to display both STX collateral type. We just need the A type as this is the most common. And then, let users pick between A or B as a second step. Could you confirm @philipdesmedt?

philipdesmedt:
@nieldeckx @Macxim don't have a strong opinion here, but perhaps it's less confusing if we only show STX-A first... or not. Not sure really. I think this was Pieter's suggestion tho?

Either way works for me.

Macxim:
Yes, it was Pieter's idea indeed. Given the B-type is not really widely used.
Also once we have atAlex, it's gonna look weird to have four cards (we could still have those 3 in the meantime though).

nieldeckx:
Made sure the page loads when user is not connected. Have also removed the STX-B card again.

I think Pieter's suggestion was to have 2 STX cards, but not for the 2 collateral types. Instead have 1 card for a stacking vault, and 1 for a non-stacking vault. So it's immediately clear to the user that in 1 case he earns a yield, in the other case his tokens remain liquid. 

Maybe something to discuss on Monday :)

philipdesmedt:
@nieldeckx 👍

nieldeckx:
@Macxim I guess this can be merged or something left to do? :)

nieldeckx (approved):

Macxim:
Yes, feel free to merge if it's good for you. 👍

PR #513, opened 2022-08-09
feat: new position cards

Macxim:
Screen Shot 2022-08-09 at 12 24 38 PM


nieldeckx (changes requested):
Looks good man! 

- Can you solve the merge conflict?
- Can you post a screenshot of how the card looks on a vault in danger and liquidated vault?

- What's the reasoning behind removing the vault ID?
- Card is a bit empty. We could fit 3 on one row?
- These cards are aligned to the left. The collateral types are centered. Would align all to the left to make it cleaner.
- Cards look more modern compared to tables. We might want to have cards for prices at the bottom as well? :P

Macxim:
- Solved

|Warning|Danger|
|---|---|
|Screen Shot 2022-08-09 at 5 04 46 PM|Screen Shot 2022-08-09 at 5 02 50 PM|

I was thinking about adding a red border or shadow or something to draw more attention on it when it's in danger.

- The id doesn't bring much value here IMO. However it's still available on the Manage page
- I could try with a full-width horizontal version
- The collateral types are centered for now but they won't be for long 😏 
- We actually have cards on mobile for prices

nieldeckx (approved):

Macxim:
> I was thinking about adding a red border or shadow or something to draw more attention on it when it's in danger.
@nieldeckx done, please check.

Macxim:
Is there a way to avoid that red flickering? Doesn't look very good.
I'm guessing it's because values haven't arrived yet.

https://user-images.githubusercontent.com/1909957/184343362-2275048c-e8d7-4550-9e7d-27a8cb44bc56.mov

Macxim:
fyi, I rebased with `master` to fix conflicts, hope I didn't mess up anything.

philipdesmedt:
@Macxim you could avoid it by setting the default class to green, or to add an additional loading state until that specific piece of data is loaded

Macxim:
@philipdesmedt done ✅

nieldeckx (approved):
Great job once again! I like the `health-status` component, clean code. :)

PR #523, opened 2022-10-31
fix: require 0.25% of supply to start a governance vote (250K DIKO)

philipdesmedt:


nieldeckx (approved):
Good catch

philipdesmedt:
@nieldeckx could you give this another good look? I updated the tests

Once good - I think we can start a gov proposal to upgrade to v4

nieldeckx:
@philipdesmedt updated the test to check if user can propose with exactly 0.25% and not with 1 DIKO less. It works, so we can start a proposal.

PR #524, opened 2022-11-03
feat: multisig oracle updates

nieldeckx:


clarity/contracts/arkadiko-oracle-v1-1.clar
```diff
@@ -43,3 +53,102 @@
...
+(define-public (set-minimum-valid-signers (minimum uint))
+  (begin
+    (asserts! (is-eq tx-sender (contract-call? .arkadiko-dao get-dao-owner)) (err ERR-NOT-AUTHORIZED))
+    
+    (var-set minimum-valid-signers minimum)
+    (ok true)
+  )
+)
+
+(define-public (set-trusted-oracle (pubkey (buff 33)) (trusted bool))
```
philipdesmedt, line 147: Q: could we remove this method?

I know this reduces flexibility significantly, but I would hardcode the pubkeys of trusted oracles on deployment and if we need to update them, we would update the oracle contract through governance. We could also do this in a v3 if we know this version works well

the reason I'm asking is because this defeats the whole purpose of not having a single point of failure.. if the DAO owner keys would somehow be compromised, our oracle is also screwed... and if we remove this method, we at least keep our oracle safe?

clarity/tests/arkadiko-oracle-v1-1_test.ts
```diff
@@ -6,6 +6,161 @@ import {
   types,
 } from "https://deno.land/x/clarinet/index.ts";
 
+
+// TODO: move to utils file
+function hexToBytes(hex: string) {
+	return hexToBytesHelper(hex.substring(0, 2) === '0x' ? hex.substring(2) : hex);
```
philipdesmedt, line 12: NTH: fix indentation and move to utils

philipdesmedt (changes requested):
I need to review the actual logic in more depth but just a comment now

clarity/contracts/arkadiko-oracle-v1-1.clar
nieldeckx: The DAO owner keys are not stored an a server, unlike the oracle keys.

Not sure if this method really matters for now because:
- There is another method "update-price-owner" to allow the DAO to update prices in case something goes wrong. Would like to keep this method for now until this new system has proven itself. So if the DAO keys are compromised, it's easier for the attacker to use the method "update-price-owner".
- If someone would get his hands on the DAO keys, he can get all liquidity from swap and reserves and all tokens in that wallet. Not sure if at that point the oracle still matters :P

nieldeckx:
@philipdesmedt this is done now. 

I think next we need to:
- Deploy the new oracle SC (v2)
- Create a new codebase for the oracle nodes, using the 'sign-message.js' script. 
- Deploy the oracle nodes on different servers with different private keys
- Let the new system run for a couple of weeks
- Once confident that it works, we can add a governance proposal to update

clarity/contracts/arkadiko-oracle-v1-1.clar
```diff
@@ -18,28 +30,225 @@
...
+)
+
+;; @desc update token price as multisig node
+;; @param block; the block on which the message was created
+;; @param token-id; token ID to update price for
+;; @param price; price value
+;; @param decimals; amount of decimals
+;; @param signatures; list of signatures from oracle nodes
+;; @post bool; returns true if successful
+(define-public (update-price-multi (block uint) (token-id uint) (price uint) (decimals uint) (signatures (list 10 (buff 65))))
```
philipdesmedt, line 96: Q: can you explain how this is being called by 3 separate nodes and only then updated? I don't get this? It seems like this method only needs to be called once with the right signatures? How is it possible to spread this among servers then?

philipdesmedt (changes requested):

clarity/contracts/arkadiko-oracle-v1-1.clar
nieldeckx: A: Correct, it needs to be called once with at least 3 correct signatures to update the price. 

The idea is that we will run 5 different nodes and they each check if the price needs to be updated based on the last  update block (and if other nodes have not pushed an update in the mempool). If an update is needed the node will fetch the price from it's source (we have 5 different sources for STX and BTC, the other tokens are fetched directly from the AMMs on chain). The node will ask the other nodes to sign the price data (block, token ID, price, decimals). The other nodes will only do this if the provided input is correct (price is within certain range of the price of their source). If the node managed to get 3 or more signatures, it pushes the transaction on chain.

I have shared the oracle node repo as well.

philipdesmedt: @nieldeckx I see, thanks, so a lot of logic is ran off-chain.

Can you share the link of the Oracle node repo here? Can't see or find it

nieldeckx: @philipdesmedt https://github.com/arkadiko-dao/oracle-node (private repo for now)

PR #518, opened 2022-08-30
feat: liquidation UI contract (dynamic tokens + lower cost)

nieldeckx:
- Dynamic token when claiming
- Option to not increate last reward ID

PR #528, opened 2022-11-13
fix: allow claiming atALEX liquidation rewards

philipdesmedt:


nieldeckx:
@philipdesmedt going to close this PR, as this is fixed in https://github.com/arkadiko-dao/arkadiko/pull/518

PR #529, opened 2022-11-15
fix: auction-engine oracle decimals

nieldeckx:
@philipdesmedt`arkadiko-auction-engine-v4-2_test` L100 & L109 shows the discounted price is wrong when xBTC price has 8 decimals in the oracle.

I calculated the output of method `get-collateral-discounted-price` manually:
Screenshot 2022-11-15 at 19 43 31

The method `get-collateral-discounted-price` takes into account the decimals. However, I don't think that's needed as the discounted price is just 90% of the current price.


nieldeckx:
@philipdesmedt i think the issue is fixed, added tests for STX, xBTC and atALEX collateral types with amount of decimals we use in prod oracle. 
To fix we don't take the decimals into account when calculating the discounted price, as this is just a percentage of the price set in the oracle. But we do take decimals into account when calculating the amount of collateral to sell, as the collateral amount set when creating a vault is also taking decimals into account.

clarity/contracts/arkadiko-stacker-payer-v3-3.clar
```diff
@@ -0,0 +1,143 @@
...
+(define-private (min-of (i1 uint) (i2 uint))
+  (if (< i1 i2)
+      i1
+      i2))
+
+
+;; Initialization
+(begin
+  ;; TODO: update for mainnet
+  (map-set vaults-redeemed { vault-id: u409 } { redeemed: true })
```
philipdesmedt, line 141: Q: do we still need all these vaults? That list becomes pretty long and hard to fetch, but I can try 😅

philipdesmedt (approved):
One question.

Then we need to deploy new auction engine and stacker payer, and start a gov vote?

nieldeckx:
**Then we need to deploy new auction engine and stacker payer, and start a gov vote?**
Yep. See script `governance-update-auction-engine` to start the gov proposal

clarity/contracts/arkadiko-stacker-payer-v3-3.clar
nieldeckx: A: now that i look at it again i don't think so. Because in the method `add-stx-redeemable` we get auctions from the newest auction-engine, so it's not possible to add redeemable stx from auctions opened by the previous auction-engine.
I guess we just need to set `stx-redeemable` properly when deploying.

PR #533, opened 2023-01-13
feat: propose new emission sinks after stable swap launch

philipdesmedt:


nieldeckx:
@philipdesmedt to complete this I think we need to:
- Create a new pool contract like `arkadiko-stake-pool-wstx-usda-v1-1` but with new LP token
- Update `arkadiko-dao` so new pool contract can mint and burn
- Update `arkadiko-stake-registry-v1-1` with new emissions and add new pool contract
- Update rewards on `arkadiko-alex-dual-yield-v1-1` using method `update-rewards-per-cycle`
- Update website to show new LP stake pool

& update the new tokenomics branch to include the new LP token

nieldeckx:
Added new stake pool and scripts. Clarinet tests will fail as the ALEX LP token contract is not found. Once it is deployed, we can add it to the deployment plan.

**To make the updates:**
- Depoy `arkadiko-stake-pool-xusd-usda-v1-1`
- Run the 4 new scripts: `new-emissions-1.js`, `new-emissions-2.js`,...

PR #536, opened 2023-02-28
feat: add additional LP token for xUSD/USDA stable swap

philipdesmedt:


philipdesmedt:
@nieldeckx the pool has been created

we still need to run the following methods
```
set-fee-rate-x
set-fee-rate-y
set-oracle-enabled
set-oracle-average
set-start-block
```

https://explorer.alexlab.co/txid/0x42b5584193460cf5f56a54fe06f039bfd127ee767452fbdf37ce4b2b51a92aa2?chain=mainnet

PR #540, opened 2023-03-22
chore: update API urls to Hiro

philipdesmedt:
See https://forum.stacks.org/t/domain-migrations-for-hiro-hosted-services/14609


Macxim (approved):

nieldeckx:
@philipdesmedt have updated the explorer URLs to avoid the redirect.

PR #550, opened 2023-05-20
Drop stacking.club API in favor of Hiro API

xbtcbridge:
The stacking.club API has been down for some time, so this PR moves the calculation of cycle starts and ends to the Hiro API. 

Note I don't actually know where to find other truth data to compare to, so I'm worried I might be off by one on the start or end blocks. So would appreciate someone double-checking this.


philipdesmedt (approved):
This is amazing! Thank you very much!

PR #552, opened 2023-06-23
feat: sorted vaults (WIP)

nieldeckx:


nieldeckx:
@philipdesmedt POC to keep a sorted list of vaults on chain. 

The same list should be kept off chain as well in order to quickly find the prev and next IDs when inserting. Based on the off chain data the FE can get a hint from the API on where to insert. As the API data might be outdated, the FE uses the hint to find the actual insert location using on-chain data. 

One downside I see is when the list is altered multiple times in the same block. The prev and next IDs might not be correct anymore after the first change. I think we can solve this by checking the 5 previous and next vaults on chain based on the prev and next ID arguments, to find the actual prev and next IDs.

Haven't looked into how this can be integrated in the current system.

PR #567, opened 2024-01-28
feat: upgrade stacker payer for liquidated vaults

philipdesmedt:


philipdesmedt:
@nieldeckx review pls

main thing I changed is

1. added a method `unlock-liquidated-vault` to v3-8 so we can unlock liquidated vaults
2. added a method `get-vault-combined-unlock` to take into account previously set unlocks on v3-7 up to the end of this current pox cycle

clarity/contracts/vaults/arkadiko-stacker-payer-v3-8.clar
```diff
@@ -0,0 +1,129 @@
...
+  )
+)
+
+(define-read-only (get-vault-combined-unlock (vault-id uint))
+  (let (
+    (old-unlock (contract-call? .arkadiko-stacker-payer-v3-7 get-vault-unlock vault-id))
+    (old-unlock-burn-height (get unlocked-at-burn-height old-unlock))
+    (new-unlock (get-vault-unlock vault-id))
+  )
+    (if (<= old-unlock-burn-height u829850)
```
nieldeckx, line 30: Q: Why this block?

nieldeckx (commented):
Looks good, just one question.

clarity/contracts/vaults/arkadiko-stacker-payer-v3-8.clar
philipdesmedt: A: this is the end block of the next PoX cycle, meaning anyone who triggered an unlock on the `v3-7` contract will get this block set for them (and thus we need to respect that unlock for these users).

Users who unlock after will get a higher block, and thus will be using the `v3-8` `unlocks` map

PR #572, opened 2024-02-10
fix: nakamoto update - contracts

nieldeckx:


clarity/contracts/arkadiko-governance-v4-2.clar
```diff
@@ -40,7 +40,7 @@
 )
 
 (define-data-var governance-shutdown-activated bool false)
-(define-data-var proposal-count uint u6)
+(define-data-var proposal-count uint u0)
```
philipdesmedt, line 43: Q: shouldn't we set this to 18? (the ID of the next gov vote)

clarity/contracts/arkadiko-diko-init-v2-1.clar
```diff
@@ -0,0 +1,159 @@
...
+    (var-set founders-tokens-claimed (+ claimed-tokens amount))
+    (as-contract (contract-call? .arkadiko-token transfer amount (as-contract tx-sender) wallet none))
+  )
+)
+
+;; ---------------------------------------------------------
+;; Migrate
+;; ---------------------------------------------------------
+
+(define-public (migrate)
```
philipdesmedt, line 141: Q: would it be better to execute this on deployment (i.e. just statements at the end of this file)? This way it is ran for sure and people know we wouldn't mint founder tokens twice etc.

philipdesmedt (commented):
Need to review all these files better, it's hard to figure out the diff. I assume most contracts are redeployments of the same with updated references to the changed contracts?

clarity/contracts/arkadiko-governance-v4-2.clar
nieldeckx: A: Not needed as the FE has logic to handle this properly. Some FE logic was needed anyway, as the proposal ID was not properly set in previous governance contracts.

nieldeckx:
> Need to review all these files better, it's hard to figure out the diff. I assume most contracts are redeployments of the same with updated references to the changed contracts?

Yes, best to diff locally. Most updates are `burn-block-height` related, and references to new contracts indeed.

lito925:
Good

PR #582, opened 2024-03-24
fix: stability fees

nieldeckx:


clarity/contracts/vaults-v2/arkadiko-vaults-operations-v1-1.clar
```diff
@@ -119,12 +119,14 @@
 )
   (let (
     (owner tx-sender)
-    (nicr (/ (* collateral u100000000) debt))
-    (collateral-info (unwrap! (contract-call? vaults-tokens get-token (contract-of token)) (err ERR_UNKNOWN_TOKEN)))
     (stability-fee (try! (contract-call? vaults-helpers get-stability-fee vaults-tokens vaults-data owner (contract-of token))))
+    (new-debt (+ debt stability-fee))
+
+    (nicr (/ (* collateral u100000000) new-debt))
```
philiphacks, line 125: Q: isn't this gonna result in issues for the positioning in the off-chain vaults API? the calculation method has changed? Or is this already incorporated in the vaults API?

philiphacks (changes requested):

clarity/contracts/vaults-v2/arkadiko-vaults-operations-v1-1.clar
nieldeckx: A: The FE will need to get the current stability fees, add these to the debt amount and use the total to call the hints API. Similar to the redemption script.

clarity/contracts/vaults-v2/arkadiko-vaults-operations-v1-1.clar
```diff
@@ -135,12 +137,12 @@
     (asserts! (not (var-get shutdown-activated)) (err ERR_SHUTDOWN))
     (asserts! (is-eq (get status vault) STATUS_ACTIVE) (err ERR_WRONG_STATUS))
     (asserts! (get valid coll-to-debt) (err ERR_INVALID_RATIO))
-    (asserts! (< (+ (- total-debt (get debt vault)) debt) (get max-debt collateral-info)) (err ERR_MAX_DEBT_REACHED))
-    (asserts! (>= debt (get vault-min-debt collateral-info)) (err ERR_MIN_DEBT))
+    (asserts! (< (+ (- total-debt (get debt vault)) new-debt) (get max-debt collateral-info)) (err ERR_MAX_DEBT_REACHED))
```
philiphacks, line 140: Q/REQ: This is potentially an issue. We set caps and these can be hit, if we hit a cap, no one can update their vault anymore (except burning debt) cause the stability fees are constantly raising the debt amounts. Any thoughts on how to fix this?

philiphacks (changes requested):

clarity/contracts/vaults-v2/arkadiko-vaults-operations-v1-1.clar
nieldeckx: A: Right. We can just use `debt` instead of `new-debt` in this calculation to not take stability fees into account for the debt ceiling.

philiphacks: Yeah indeed. Simply using `debt` instead of `new-debt` is the easy fix :)

philiphacks (changes requested):

PR #588, opened 2024-04-23
feat: add DIKO v2 staking pool without cooldown

philiphacks:
Instead of waiting 10 days, this update allows unstaking immediately in the next block.

TODO after deploy & gov vote:

- Set new DIKO pool in `pools-data-map` on `arkadiko-stake-registry-v2-1`
- Migrate staked DIKO from `arkadiko-stake-pool-diko-v1-4` to new pool
- Deploy and use `stake/arkadiko-ui-stake-v1-5.clar`


nieldeckx (approved):
Looks good!

PR #598, opened 2024-09-25
(WIP) feat: add redemptions page with info and least healthy vault

philiphacks:


philiphacks:
@nieldeckx can you review pls?

web/components/redeem-vault.tsx
```diff
@@ -0,0 +1,151 @@
...
+      makeStandardFungiblePostCondition(
+        stxAddress || '',
+        FungibleConditionCode.Equal,
+        amount.value,
+        createAssetInfo(contractAddress, 'usda-token', 'usda')
+      ),
+    ];
+
+    const vault = collateralToRedeem === 'stSTX' ? stStxVault : collateralToRedeem === 'xBTC' ? xBtcVault : stxVault;
+    const remainingDebt = Math.max(vault.debt - redeemAmount, 0);
```
nieldeckx, line 71: REQ: when redeeming, the stability fee is added to the vault debt.
So this needs te become something like `(vault-debt + vault['stability-fee']) - redeemAmount`

web/components/redeem-vault.tsx
```diff
@@ -0,0 +1,151 @@
...
+
+    const BASE_URL = process.env.HINT_API_URL;
+    const url = BASE_URL + `?owner=${vault['owner']}&token=${vault['token']}&collateral=${vault['collateral']}&debt=${remainingDebt}`;
```
nieldeckx, line 74: REQ: the collateral param needs to be the collateral amount left after redeeming.

nieldeckx (changes requested):

web/components/redeem-vault.tsx
philiphacks: @nieldeckx does this need to be exact? what would be the best way to calculate this?

nieldeckx: First calculate the amount of collateral that will be redeemed based on the debt that is payed off, and subtract this from the current collateral.
So something like: `vault['collateral'] - (redeemAmount / tokenPrice)`

philiphacks:
@nieldeckx fixed your comments, can you review again pls?

nieldeckx (approved):
LGTM

PR #607, opened 2025-04-15
feat: oracle-v2-4

nieldeckx:


clarity/contracts/arkadiko-oracle-v2-4.clar
```diff
@@ -0,0 +1,252 @@
...
+;; Price Helpers
+;; ---------------------------------------------------------
+
+(define-public (get-pyth-info (token (string-ascii 12)))
+  (let (
+    (price-info (map-get? prices-pyth { token: token }))
+  )
+    (if (is-some price-info)
+      (let (
+        (price (try! (get-price-pyth (get feed-id (unwrap-panic price-info)))))
```
philiphacks, line 98: Q: since on line 67, we check `(is-ok pyth-result)`, should we do `unwrap-panic` here? I don't know if this can ever panic, but it would make the `fetch-price` method fail? same for logic in the DIA method and custom method

clarity/contracts/arkadiko-oracle-v2-4.clar
```diff
@@ -0,0 +1,252 @@
...
+
+(define-public (get-price-pyth (price-feed-id (buff 32)))
+  (let (
+
+    ;; TODO: Update for mainnet
+    ;; (storage 'SP3R4F6C1J3JQWWCVZ3S7FRRYPMYG6ZW6RZK31FXY.pyth-storage-v3)
+    ;; (oracle-data (try! (contract-call? 'SP3R4F6C1J3JQWWCVZ3S7FRRYPMYG6ZW6RZK31FXY.pyth-oracle-v3 get-price price-feed-id storage)))
+
+    (storage .pyth-storage-v3)
+    (oracle-data (try! (contract-call? .pyth-oracle-v3 get-price price-feed-id storage)))
```
philiphacks, line 169: Q: I thought pyth was a push oracle, meaning we are responsible for updating the prices (which is often done as part of a user action), is this correct? Or are we just going to rely on other protocols keeping the price up to date (e.g. Zest/Granite)?

clarity/contracts/arkadiko-oracle-v2-4.clar
```diff
@@ -0,0 +1,252 @@
...
+  )
+    (ok (to-uint (get price oracle-data)))
+  )
+)
+
+(define-public (get-price-dia (key (string-ascii 32)))
+  (let (
+    ;; TODO: Update for mainnet
+    ;; (oracle-data (unwrap-panic (contract-call? 'SP1G48FZ4Y7JY8G2Z0N51QTCYGBQ6F4J43J77BQC0.dia-oracle get-value key)))
+    (oracle-data (unwrap-panic (contract-call? .dia-oracle get-value key)))
```
philiphacks, line 179: Q: do we have a list of the keys for DIA, and also for Pyth which `price-feed-id` values we can pass? would be good to keep them as comments here perhaps

clarity/contracts/arkadiko-oracle-v2-4.clar
```diff
@@ -0,0 +1,252 @@
...
+;; ---------------------------------------------------------
+
+;; Oracle prices are standardized to 8 decimal places (100000000).
+;; For tokens with 6 decimals (like STX), we divide by 100000000 to normalize.
+;; For tokens with 8 decimals (like xBTC/sBTC), we divide by an additional 100 in the Arkadiko protocol to account for the extra precision.
+
+(define-constant decimals8 u100000000)
+
+(begin
+  (map-set prices-pyth { token: "STX" } { decimals: decimals8, feed-id: 0xec7a775f46379b5e943c3526b1c8d54cd49749176b0b98e02dde68d1bd335c17 })
```
philiphacks, line 240: NVM - I see the price feed IDs here

clarity/contracts/arkadiko-oracle-v2-4.clar
```diff
@@ -0,0 +1,252 @@
...
+  (map-set prices-pyth { token: "xSTX" } { decimals: decimals8, feed-id: 0xec7a775f46379b5e943c3526b1c8d54cd49749176b0b98e02dde68d1bd335c17 })
+
+  (map-set prices-pyth { token: "BTC" } { decimals: (* decimals8 u100), feed-id: 0xe62df6c8b4a85fe1a67db44dc12de5db330f7ac66b72dc658afedf0f4a415b43 })
+  (map-set prices-pyth { token: "xBTC" } { decimals: (* decimals8 u100), feed-id: 0xe62df6c8b4a85fe1a67db44dc12de5db330f7ac66b72dc658afedf0f4a415b43 })
+  (map-set prices-pyth { token: "sBTC" } { decimals: (* decimals8 u100), feed-id: 0xe62df6c8b4a85fe1a67db44dc12de5db330f7ac66b72dc658afedf0f4a415b43 })
+
+  (map-set prices-custom { token: "stSTX" } { decimals: decimals8, key: "stSTX/USD" })
+  (map-set prices-custom { token: "ststx-token" } { decimals: decimals8, key: "stSTX/USD" })
+
+  (map-set prices-dia { token: "DIKO" } { decimals: decimals8, key: "DIKO/USD" })
```
philiphacks, line 250: Q: how does DIA set the price of DIKO?

It's probably not important since only Zest uses it as collateral (and you cannot borrow against it afaik), but just wondering... they probably use our DEX?

clarity/contracts/arkadiko-oracle-v2-4.clar
nieldeckx: A: Correct. If we would fully implement pyth, it would mean that we push prices when a vault gets updated. As other protocols already update prices on a regular basis, I don't think it's worth the change.

clarity/contracts/arkadiko-oracle-v2-4.clar
nieldeckx: A: from what I found they use BitFlow (which aggregates all dexes on stacks). 
https://www.diadata.org/blog/post/stacks-bitcoin-l2-dia-oracles/

clarity/contracts/arkadiko-oracle-v2-4.clar
nieldeckx: A: We could do `unwrap-panic` here as well, but using `try!` makes sure the actual error is returned making it easier to debug in case something goes wrong. In both cases the `is-ok` on line 67 would fail.

philiphacks: Ok, perfect!

philiphacks:
@nieldeckx looks good to me to deploy & test for a few days

